PR #5, opened 2016-01-11
Add stack

Arguggi:


lukasmartinelli:
Thanks @Arguggi! Need to check out Stack, had alot of trouble with getting to know Cabal.

lukasmartinelli:
Building was much faster with stack!

PR #10, opened 2016-01-12
Fix README to reference piping to docker

moshen:
Piping is a little simpler than mounting a volume and also works well
for those running docker in a virtual machine.


lukasmartinelli:
Yeah. I tried piping at first (that's why there is an `-i` flag) but it didn't work out. Thanks for figuring it out.

moshen:
No problem.  Since I do most of my docker dev in a VirtualBox docker-machine, I was trying to avoid the volume mounting madness. :beers:

PR #12, opened 2016-01-13
Sort messages by line number and return an exit code

cpennington:


cpennington:
Oops. I meant to squash the commits before this was merged. But thanks for the quick response!

PR #18, opened 2016-01-20
Fix typo

pdehaan:


lukasmartinelli:
My nemesis word. Thanks.

jsoref:
The art should be updated as well...

Issue #20, opened 2016-02-01
Code Climate Engine


devth:
Would be useful to run `hadolint` as a Code Climate Engine:
http://blog.codeclimate.com/blog/2015/07/07/build-your-own-codeclimate-engine/

lukasmartinelli:
- [ ] Need to trim down the image size from 1 GB to 250MB first (but that's a good thing anyway). Will probably create a special distribution file.
- [ ] Implement a JSON formatter

lukasmartinelli:
Just remember that your initial suggestion was to use the API from the website.

 That will be much less work to integrate.

devth:
I was actually thinking of running it as an engine, but API would be nice too.

The Yetibot plugin runs CodeClimate locally, so the only way it could take advantage of hadolint is if it were packaged as an engine.

lukasmartinelli:
> The Yetibot plugin runs CodeClimate locally, so the only way it could take advantage of hadolint is if it were packaged as an engine.

Wuhuu. It is actually not that difficult to build a static Haskell binary (building it crossplatform is though).
I can now have build container that spits out the binary that I can then add to a blank image and stay under the 200MB limit. Should write a Makefile for that.

devth:
Nice! I wonder if Travis or CircleCI or a similar could help with cross-platform builds?

https://circleci.com/docs/language-haskell

lukasmartinelli:
Will take a bit longer. In order to make it work I need to implement #9 and #13 first - and file globbing as well.

First thought about just writing a wrapper script in a other language to simply transform the output into JSON and deal with all the other stuff - but that would be giving up on Haskell :grimacing:

> Nice! I wonder if Travis or CircleCI or a similar could help with cross-platform builds?

The only good crosscompile story I've seen so far is in Go. Haven't found something on the internet
to simply create binaries for OSX, Windows and Linux for Haskell on one single platform.
But you're right, I could just build a binary on the corresponding platform.

devth:
I thought Docker would remove the need to cross-compile.

dblandin:
Hey @devth @lukasmartinelli,

I'm Devon, an engineer at @codeclimate.

Awesome to see that you're considering hadolint for a Code Climate engine! We have a `Dockerfile` in nearly every codeclimate repo and would definitely benefit from this tool during our own analyses.

[...]

Issue #9, opened 2016-01-12
Display Dockerfile path when showing the results


ye:
Currently, only line numbers are printed out when there is a rule violation detected. It would be great to have the path to the Dockerfile examined printed out as well, just like most of other linters/compilers do.
That way, we can write a simple bash script to lint a whole bunch of Dockerfiles and the results would make more sense as to while violation is pertaining to which Dockerfile.
Additionally, `hadolint` perhaps can take a directory and find the Dockerfiles in it and show results altogether instead of having to run it one Dockerfile at a time.

lukasmartinelli:
> Currently, only line numbers are printed out when there is a rule violation detected. It would be great to have the path to the Dockerfile examined printed out as well, just like most of other linters/compilers do.
> That way, we can write a simple bash script to lint a whole bunch of Dockerfiles and the results would make more sense as to while violation is pertaining to which Dockerfile.
> Additionally, hadolint perhaps can take a directory and find the Dockerfiles in it and show results altogether instead of having to run it one Dockerfile at a time.

That all makes a whole lot of sense! The CLI version is very basic, I want to implement multiple formatters to support better output.

Actuall didn't think people would use the CLI version because installing something with Haskell is kind of difficult.

ye:
@lukasmartinelli since there is no GUI, how else would you expect people to use `hadolint` other than CLI? ;) In fact, you'd be surprised a lot of us on Github know how to compile code into binaries and know a thing or two about Unix and command line!

lukasmartinelli:
Fixed by #35.

PR #35, opened 2016-05-21
Display Dockerfile path with results

wfleming:
Hi @lukasmartinelli,

I'm an engineer at Code Climate & am interested in picking away at some of the features Hadolint would need to make implementing #20 easier: I think Hadolint would be a great engine for our platform.

I'm starting with a small fix for #9, so that the source file name can be carried along with the check information & displayed in the output.

### DIFF
--- a/src/Formatter.hs
+++ b/src/Formatter.hs
@@ -5,9 +5,9 @@ import Rules
 import Syntax
 
 formatCheck :: Check -> String
-formatCheck (Check metadata linenumber _) = formatPos linenumber ++ code metadata ++ " " ++ message metadata
+formatCheck (Check metadata source linenumber _) = formatPos source linenumber ++ code metadata ++ " " ++ message metadata
 
-formatPos :: Linenumber -> String
-formatPos linenumber = if linenumber >= 0
-                       then "L" ++ show linenumber ++ " "
-                       else ""
+formatPos :: Filename -> Linenumber -> String
+formatPos source linenumber = if linenumber >= 0
+                              then source ++ ":" ++ show linenumber ++ " "
+                              else source ++ " "
--- a/src/Parser.hs
+++ b/src/Parser.hs
@@ -237,7 +237,7 @@ dockerfile = many $ do
     pos <- getPosition
     i <- parseInstruction
     many eol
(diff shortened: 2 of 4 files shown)


lukasmartinelli:
> I'm an engineer at Code Climate & am interested in picking away at some of the features Hadolint would need to make implementing #20 easier: I think Hadolint would be a great engine for our platform.

Now I feel bad I haven't continued at #20 but also motivated to fix the many bugs that were filed so far.
Many thanks.



What are next steps to do? Write a JSON formatter to fulfil the codeclimate spec https://github.com/codeclimate/spec/blob/master/SPEC.md?

wfleming:
Cool, thanks!

> What are next steps to do? Write a JSON formatter to fulfil the codeclimate spec https://github.com/codeclimate/spec/blob/master/SPEC.md?

Yes, that would be an important next step. I think the other important thing we'd need to do is using/complying with the [`config.json`](https://github.com/codeclimate/spec/blob/master/SPEC.md#which-files-to-analyze) we provide to engines to filter out which files can be used for analysis. It would also be great to get to a point where the tool can detect dockerfiles by content (e.g. by looking for a first line  that starts with `FROM `), but that might be a later feature that would be good to add.

There's also some nit-picky little things that are different between our spec & what makes sense for a stand-alone CLI. For example, it's important to exit with code `0` [except in the case of fatal errors](https://github.com/codeclimate/spec/blob/master/SPEC.md#output).

You mentioned the possibility in #20 of a wrapper script, and that might be a good way to go. The JSON formatter would probably be good to keep in the main `hadolint` binary (re-parsing the output just to print it again seems error-prone & unpleasant), but things like handling `config.json` and looking for dockerfiles might be better off in the wrapper script: a number of our engines take this kind of approach. It doesn't necessarily need to mean abandoning Haskell either: you could either have  a separate repo for that wrapper, or maybe write it as a second binary target in this repo.

You might be interested in joining our [developer program](https://codeclimate.com/dev_form): we have a Slack community where members of our team & authors of our community engines can hang out and talk about the details of getting Code Climate engines shipped.

Feel free to ping me in that community to chat through more details if you want. I'd like to help with this however I can.

Issue #40, opened 2016-08-10
Allow to pass a set of rules to ignore


dokempf:
It would be nice if it would be possible to pass a set of rules that should be ignored by the linter. Compare to python, where pep8 is an accepted standard, but almost every project defines 2-3 well-thought-of exceptions. I cannot simply ignore the output, as I want to use hadolint in a CI system.

FWIW, the rule that I disagree with is DL3008.

ye:
This is a duplicate of #24

Issue #24, opened 2016-02-16
We should have an ability to ignore some rules.


niksite:
E.g. `hadolint -i --ignore=DL3008` would be much appreciated.

lukasmartinelli:
:+1:

timgriffiths:
Would be better if you could add a comment in the dockerfile directly above what you wanted to ignore so it was clear to all users. Ie
#DL3008 ignoring because abc

bootstraponline:
:+1:

jmreicha:
I'd like to see something like @timgriffiths mentioned.  I think something similar to how Shellcheck  works would be nice.

```
# hadolint disable=DL3006
FROM ubuntu
```

For example.

ye:
I, too, agree with @timgriffiths that Dockerfile line comment is a better and more flexible way to ignore rules than using command line argument(s). The difference lies in the fact that you can cherry picking the local scope where certain rule(s) should be ignored while maintaining the check on other places. If we are implementing a command line argument, it's less flexible as the rule ignore will be applied globally.

This line comment ignore rule practice is also used in other linters like [pylint](https://docs.pylint.org/en/latest/faq.html#is-it-possible-to-locally-disable-a-particular-message)

lukasmartinelli:
> I, too, agree with @timgriffiths that Dockerfile line comment is a better and more flexible way to ignore rules than using command line argument(s). The difference lies in the fact that you can cherry picking the local scope where certain rule(s) should be ignored while maintaining the check on other places. If we are implementing a command line argument, it's less flexible as the rule ignore will be applied globally.

I like the comment version more as well. It is a bit more tricky to implement since it is then part of the AST and it is the first time that two Dockerfile instructions would relate to each other. The `RUN` instruction would only **not** be tested against the rule if a `COMMENT` instruction one line above exists and has it disabled.

To not get lost with this I will start with a** command line version** since with this I can just filter out all lint results from that rule. Build an MVP first 😁

ye:
Excited for v1.1, can't wait! 

Any indicator of timeframe for v1.1 to be released @lukasmartinelli ? Without the ignore rule feature, my CI build is currently failing... :(

![screen shot 2016-08-18 at 1 41 16 pm](https://cloud.githubusercontent.com/assets/133924/17784226/b76241f4-6549-11e6-9128-ed89942c2852.png)

lukasmartinelli:
@ye Sorry never followed up - this should be included in the latest releases
[...]

PR #44, opened 2016-08-18
Better CLI interface

lukasmartinelli:
- Read from stdin will be gone
- Now provide a nice help text
- Support ignore filters #24 #40


ye:
> * Read from stdin will be gone

Did this mean that Unix pipeline usage of `hadolint` will be broken? Oh no!

lukasmartinelli:
> Did this mean that Unix pipeline usage of hadolint will be broken? Oh no!

You can still pipe stuff into hadolint since stdin is also just a file.

```
cat Dockerfile | hadolint /dev/stdin
```

ye:
Good to know that stdin pipes can still be used albeit it's a bit awkward now in the current syntax. A tiny suggestion: can it interpret the dash `-` as the equivalent of `/dev/stdin` like most other unix tools do?

lukasmartinelli:
> Good to know that stdin pipes can still be used albeit it's a bit awkward now in the current syntax. A tiny suggestion: can it interpret the dash - as the equivalent of /dev/stdin like most other unix tools do?
>  👍 1

Good idea. Low hanging fruit. Done in https://github.com/lukasmartinelli/hadolint/commit/e337f71657f23f5cec36e64686becf671eafa3b6

Pattern matching to the rescue.

```haskell
parseFilename :: String -> String
parseFilename "-" = "/dev/stdin"
parseFilename s = s
```

PR #45, opened 2016-08-18
JSON Output from CLI

lukasmartinelli:
This supports JSON output compatible with Codeclimate #20.
I now upload builds via Linux and will provide binaries for hadolint that can then be used inside a slim Docker image (sadly I cannot create a completly static library so perhaps it doesn't work on all Linux machines - but in Docker we can guarantee it this way).

Code climate guys @dblandin @wfleming @mrb what else do we need that it runs on  Code climate?

I think some stuff can be added as wrapper script around it. It will now simply emit a JSON like this given a Dockerfile.

``` javascript
{
  "location": {
    "path": "Dockerfile",
    "lines": {
      "begin": 4,
      "end": 4
    }
  },
  "severity": "normal",
  "check_name": "DL3008",
  "categories": [
    "Compatibility"
  ],
  "content": "https://github.com/lukasmartinelli/hadolint/wiki/DL3008",
  "type": "issue",
  "description": "Pin versions in apt get install. Instead of `apt-get install <package>` use `apt-get install <package>=<version>`"
}
```

And put each issue on a separate line like this.

``` bash
{"location":{"path":"Dockerfile","lines":{"begin":1,"end":1}},"severity":"normal","check_name":"DL3006","categories":["Compatibility"],"content":"https://github.com/lukasmartinelli/hadolint/wiki/DL3006","type":"issue","description":"Always tag the version of an image explicitly."}
{"location":{"path":"Dockerfile","lines":{"begin":4,"end":4}},"severity":"normal","check_name":"DL3008","categories":["Compatibility"],"content":"https://github.com/lukasmartinelli/hadolint/wiki/DL3008","type":"issue","description":"Pin versions in apt get install. Instead of `apt-get install <package>` use `apt-get install <package>=<version>`"}
```

Format code is in `src/Hadolint/Format.hs`.


src/Hadolint/Format.hs
```diff
@@ -0,0 +1,45 @@
...
+categorize "DL3004" = "Security"
+categorize _ = "Compatibility"
+
+-- | Emit Code Climate compatible JSON structure of check
+-- https://github.com/codeclimate/spec/blob/master/SPEC.md
+instance ToJSON Check where
+    toJSON (Check m fname lineno success ) = object [ "type" .= T.pack "issue"
+                      , "check_name" .= T.pack (code m)
+                      , "description" .= T.pack (message m)
+                      , "content" .= ("https://github.com/lukasmartinelli/hadolint/wiki/" ++ code m)
```
wfleming: This is a fairly common area of confusion with the spec: the `content` key should have an object with a `body` key, not just a string. e.g. `{ "content": { "body": "https://github.com/lukasmartinelli/hadolint/wiki/foo"} }`.

https://github.com/codeclimate/spec/blob/master/SPEC.md#contents

lukasmartinelli: > This is a fairly common area of confusion with the spec: the content key should have an object with a body key, not just a string. e.g. { "content": { "body": "https://github.com/lukasmartinelli/hadolint/wiki/foo"} }.

Thanks. Added it.

wfleming:
> And put each issue on a separate line like this.

I may be missing something in the code, but am I correct that the issues are only separated by `"\n"`? The spec [expects issues to be separated by a null byte](https://github.com/codeclimate/spec/blob/master/SPEC.md#output), so you'll want to add that as well. (It's fine to keep the newline as well for human readability, if you want to.) Apologies if there is a null byte & I just missed it in the code.

wfleming:
This is looking great, @lukasmartinelli! 

> I think some stuff can be added as wrapper script around it

I think that makes sense. At this point I think the missing bit of functional work we'd like to do here, is [making use of the `config.json`](https://github.com/codeclimate/spec/blob/master/SPEC.md#which-files-to-analyze). Even though Hadolint by default only looks at `Dockerfile`, the spec still expects the engine to do a quick sanity check that that file hasn't been excluded (other "single file" engines do this as well). A wrapper script would be a sensible place to do this: lots of engines end up with a small wrapper script that takes care of this & any other Code Climate-specific logic. [Brakeman has a good example](https://github.com/presidentbeef/brakeman/blob/master/bin/codeclimate-brakeman) you might find useful.

A nice-to-have that the wrapper script could also take care of would be configuration of which files to analyze: `Hadolint` can take the file to analyze as a command line argument, but our platform doesn't have a generic way of specifying command line arguments right now. If you wanted to support this functionality for repos with multiple dockerfiles, you could support a key in the user's `.codeclimate.yml` which then gets written into the `config.json` you'll end up reading from. E.g. the YAML could have:

```
engines:
  hadolint
    enabled: true
    config:
      files: ["Dockerfile", "OtherDockerfile", "ThirdDockerfile"]
```

So your wrapper script would basically be responsible for figuring out which files should get analyzed (if any), and then run the Hadolint binary against those files in `/code` in the docker image with the JSON output.

I'm really excited to see this move forward. GitHub comments aren't always the best place to talk through the details of things like implementing the wrapper script & then testing the built engine image using our [CLI](https://github.com/codeclimate/codeclimate). Feel free to email me at will@codeclimate.com if you'd like to chat through implementation details more!

lukasmartinelli:
Should we still support a JSON output or close this down?

dblandin:
I think adding a wrapper script which checks the `/config.json` file for file inclusion is the only blocker on Code Climate's end. We'd love to include it within Code Cimate's engine platform with that in place.

zemanlx:
@lukasmartinelli It would be nice to have a `JSON` output. I believe it would be useful for more people. Do you plan to rebase this PR?

lukasmartinelli:
> @lukasmartinelli It would be nice to have a JSON output. I believe it would be useful for more people. Do you plan to rebase this PR?

Let's not close this. Only need a long 🛩  trip and write some haskell again :)

lorenzo:
@lukasmartinelli do you need help to finish this?

lukasmartinelli:
> @lukasmartinelli do you need help to finish this?

Of course! I would love for someone to take it on for basic JSON output so machines can parse it.

zemanlx:
It would be nice to have it compatible with shellcheck output options/format.

lorenzo:
I'll take a look at this on the weekend

lorenzo:
closing in favour of #164

Issue #36, opened 2016-05-23
PIP pinning failures


hsmade:
I found 2 occurrences where I get a false negative on PIP pinning:

RUN pip install git+https://github.com/rtfd/readthedocs-sphinx-ext.git@0.6-alpha#egg=readthedocs-sphinx-ext

RUN pip install 'alabaster>=0.7,<0.8,!=0.7.5'

hsmade:
--- a/src/Rules.hs
+++ b/src/Rules.hs
@@ -244,7 +244,7 @@ pipVersionPinned = instructionRule code severity message check
           check (Run args) = not (isPipInstall args && not (isRecursiveInstall args)) ||
                                 all versionFixed (packages args)
           check _ = True
-          versionFixed package = "==" `isInfixOf` package
+          versionFixed package = "==" `isInfixOf` package || ("git+http" `isInfixOf` package && "@" `isInfixOf` package)
           packages :: [String] -> [String]
           packages args = concat [drop 2 cmd | cmd <- bashCommands args, isPipInstall cmd]
           isPipInstall cmd = ["pip", "install"] `isInfixOf` cmd


(sadly, my current proxy settings don't allow for a fork, commit, push, pull request ;))

PR #46, opened 2016-08-18
Fix pip pinning

lukasmartinelli:
According to #36.

@hsmade Can you check the regression tests in `test/Spec.hs` whether that is correct like that?


hsmade:
Looks good to me!

Issue #34, opened 2016-05-19
Add more archive suffixes for add vs copy check


hsmade:
https://github.com/lukasmartinelli/hadolint/blob/f951912765abe12e70372a56fde568c5e252febc/src/Rules.hs#L271

Could you please add .tgz ?

hsmade:
--- a/src/Rules.hs
+++ b/src/Rules.hs
@@ -270,7 +270,7 @@ aptGetNoRecommends = instructionRule code severity message check
           hasNoRecommendsOption cmd = ["--no-install-recommends"] `isInfixOf` cmd
 
 isArchive :: String -> Bool
-isArchive path =  True `elem` [ftype `isSuffixOf` path | ftype <- [".tar", ".gz", ".bz2", ".xz"]]
+isArchive path =  True `elem` [ftype `isSuffixOf` path | ftype <- [".tar", ".gz", ".bz2", ".xz", ".tgz"]]
 isUrl :: String -> Bool
 isUrl path = True `elem` [proto `isPrefixOf` path | proto <- ["https://", "http://"]]
 copyInsteadAdd = instructionRule code severity message check

PR #47, opened 2016-08-18
Add more archive types #34

lukasmartinelli:
@hsmade More suggestions for archive files?

### DIFF
--- a/src/Hadolint/Rules.hs
+++ b/src/Hadolint/Rules.hs
@@ -270,7 +270,7 @@ aptGetNoRecommends = instructionRule code severity message check
           hasNoRecommendsOption cmd = ["--no-install-recommends"] `isInfixOf` cmd
 
 isArchive :: String -> Bool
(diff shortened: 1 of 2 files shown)


hsmade:
I can't find the exact code in docker where we can see what it supports..
The reference for dockerfile says: identity, gzip, bzip2 or xz
Somewhere in the code, I found these. But it's not related.
(tar|TAR)(.gz|.GZ|.Z|.bz2|.lzma|.xz|)|(tbz|tgz|txz))

lukasmartinelli:
> I can't find the exact code in docker where we can see what it supports..
> The reference for dockerfile says: identity, gzip, bzip2 or xz
> Somewhere in the code, I found these. But it's not related.
> (tar|TAR)(.gz|.GZ|.Z|.bz2|.lzma|.xz|)|(tbz|tgz|txz))

Ok thanks for the research. So let's leave it at `".tar", ".gz", ".bz2", ".xz", ".zip", ".tgz"` for now.

PR #60, opened 2016-09-06
Add support for parsing SHELL instruction

backus:
Let me know if I missed something. This is basically the first haskell I've written

### DIFF
--- a/src/Hadolint/Parser.hs
+++ b/src/Hadolint/Parser.hs
@@ -59,6 +59,12 @@ cmd = do
   args <- arguments
   return $ Cmd args
 
+shell :: Parser Instruction
+shell = do
+  reserved "SHELL"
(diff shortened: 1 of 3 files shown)


backus:
I also checked to confirm that some of the dockerfiles in the repos the integration tests pull down do indeed include `SHELL` instruction usage. For example: https://github.com/docker-library/mongo/blob/master/3.0/windows/windowsservercore/Dockerfile#L3

backus:
@lukasmartinelli I think this is ready for review

lukasmartinelli:
> Let me know if I missed something. This is basically the first haskell I've written

Awesome! Pretty good for the first Haskell program! Looks elegant.

backus:
@lukasmartinelli :smile:

Issue #59, opened 2016-09-06
Recommend `SHELL` over `RUN ln -sfv /bin/bash /bin/sh`


backus:
Extracted rule suggestion from https://github.com/lukasmartinelli/hadolint/issues/58. Basically, instead of writing 

```
RUN ln -sfv /bin/bash /bin/sh
```

You should write

```
SHELL ["/bin/bash", "-c"]
```

This `SHELL` instruction is new in Docker 1.12: https://docs.docker.com/engine/reference/builder/#/shell

PR #61, opened 2016-09-06
Add "use SHELL" rule for ln with /bin/sh

backus:
Addresses #59.

* I'm not sure how the numbering is decided for the rules so I just went with `DL4005` because it was the first number available after the highest `DL` number.

* There might be a better way to detect this shell symlinking behavior but right now I just assert that the `RUN` command is using `ln` and one of the arguments is `/bin/sh`. Open to suggestions for how to do this better. I'm not sure if this behaves well with chained commands for example (like `ln -s foo bar && something_with /bin/sh`)

* This doesn't handle anything for windows users. Noting this given that the docker reference recommends the `SHELL` instruction for windows users who prefix everything with `powershell -command ...`

### DIFF
--- a/src/Hadolint/Rules.hs
+++ b/src/Hadolint/Rules.hs
@@ -77,6 +77,7 @@ rules = [ absoluteWorkdir
         , hasMaintainer
         , multipleCmds
         , multipleEntrypoints
+        , useShell
         ]
 
 commentMetadata :: ShellCheck.Interface.Comment -> Metadata
@@ -292,3 +293,11 @@ copyInsteadAdd = instructionRule code severity message check
           message = "Use COPY instead of ADD for files and folders"
           check (Add src _) = isArchive src || isUrl src
           check _ = True
+
+useShell = instructionRule code severity message check
+    where code = "DL4005"
+          severity = WarningC
+          message = "Use SHELL to change the default shell"
+          check (Run args) = not $ any shellSymlink (bashCommands args)
+          check _ = True
+          shellSymlink (args) = usingProgram "ln" args && isInfixOf ["/bin/sh"] args
--- a/test/Spec.hs
+++ b/test/Spec.hs
@@ -195,6 +195,11 @@ main = hspec $ do
     it "pip install lower bound" $ ruleCatchesNot pipVersionPinned "RUN pip install 'alabaster<0.7'"
     it "pip install excluded version" $ ruleCatchesNot pipVersionPinned "RUN pip install 'alabaster!=0.7'"
 
+  describe "use SHELL" $ do
+    it "RUN ln" $ ruleCatches useShell "RUN ln -sfv /bin/bash /bin/sh"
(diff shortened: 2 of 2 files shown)


lukasmartinelli:
Great. Your contributions make my day 😁 

> I'm not sure how the numbering is decided for the rules so I just went with DL4005 because it was the first number available after the highest DL number.

That's alright.

> There might be a better way to detect this shell symlinking behavior but right now I just assert that the RUN command is using ln and one of the arguments is /bin/sh. Open to suggestions for how to do this better. I'm not sure if this behaves well with chained commands for example (like ln -s foo bar && something_with /bin/sh)

Write another spec test to assert that.
You use `usingProgram` which is using `bashCommands` to split the commands by `&&`.
But `usingProgram` will only assert that any of the commands is the ln command.

What you want is to find a `ln` command and then run the bash check.


So perhaps a variant of that (perhaps you get something better):

1. get all commands
2. map over a function that checks
  1. Is it not `ln` or is it `ln` and does it not contain `/sh/`
3. ensure that the result of the check function for all bash commands are true (with haskell `and`)


> This doesn't handle anything for windows users. Noting this given that the docker reference recommends the SHELL instruction for windows users who prefix everything with powershell -command ...

I don't think hadolint should handle Windows things - since it relies on Shellcheck to do bash handling.

backus:
👍 @lukasmartinelli I'll look into your suggestions and ping you when I've pushed the changes.

backus:
@lukasmartinelli I updated the check with the definition `not $ any shellSymlink (bashCommands args)` which I think covers the case of multiple bash commands chained together

backus:
@lukasmartinelli this should be ready for review now

lukasmartinelli:
> @lukasmartinelli I updated the check with the definition not $ any shellSymlink (bashCommands args) which I think covers the case of multiple bash commands chained together. I included a test for this to confirm

Perfect! Impressed how well you write that Haskell!

lukasmartinelli:
What were problems to get started with development of hadolint?
So in the future I can ease those barriers.
Haskell installation?
Stack build?

backus:
@lukasmartinelli I think the main thing was the README saying to do `cabal init` and `cabal install --only-dependencies`. I also had some other stale dependencies locally but once I updated to GHC 7.10.3 I was able to do `stack build` successfully.

lukasmartinelli:
I gave you push access to the repository so you can create branches and collaborate directly on this repo. Thanks again for the help so far!

> @lukasmartinelli I think the main thing was the README saying to do cabal init and cabal install --only-dependencies. I also had some other stale dependencies locally but once I updated to GHC 7.10.3 I was able to do stack build successfully.

Switched it to stack instructions.

backus:
@lukasmartinelli I'm happy to help. Your tool helped accelerate my knowledge of Docker quite a bit. 
 :smile:

Appreciate you adding me as a collaborator. I don't have too much else in mind (besides #62 and adding that wiki entry for `SHELL`) for adding to the project right now but I'll try to propose and implement other additions if they come to mind.

PR #92, opened 2017-09-04
Misc updates

j3p0uk:
Newer version of ShellCheck needed updates, and for some reason during build I got version 0.0.1 of optparse-applicative , so set a minimum that works.  These were during a cabal build of hadolint.


zemanlx:
@j3p0uk We are now on 9.10 (offers `optparse-applicative-0.13.2.0`) and use ShellCheck 0.4.6, therefore, I am closing this PR. Thank you for your contribution.

PR #100, opened 2017-11-02
Build static binaries ...

zemanlx:
... if it is an option. As I have learnt that [OS X does not support static binaries](https://developer.apple.com/library/content/qa/qa1118) for `c`. Linux and Windows builds are now build with `--static` option. 

Linux builds are static
```bash
$ ldd .stack-work/dist/x86_64-linux/*/build/hadolint/hadolint                                                                    
.stack-work/dist/x86_64-linux/Cabal-1.24.2.0/build/hadolint/hadolint:
        not a dynamic executable
.stack-work/dist/x86_64-linux/Cabal-2.0.0.2/build/hadolint/hadolint:
        not a dynamic executable
```

I do not have experience in Windows world but I have tried it to run on a fresh Windows Server 2016 installation and it was running :tada:.

@lukasmartinelli is this what you have been doing for https://github.com/lukasmartinelli/hadolint/issues/49 ?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -109,5 +109,5 @@ install:
 
 script:
 - stack --no-terminal $ARGS test --bench --no-run-benchmarks --haddock --no-haddock-deps
-- stack --no-terminal $ARGS install --ghc-options='-O0'
+- stack --no-terminal $ARGS install --ghc-options='-fPIC'
 - cp "$(which hadolint)" ./releases/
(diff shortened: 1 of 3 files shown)


lukasmartinelli:
> ... if it is an option. As I have learnt that OS X does not support static binaries for c. Linux and Windows builds are now build with --static option.

I didn't know that either.
Apart from that I guess this is good. I don't understand enough about building to say more.

Thanks for all the great PR work @zemanlx.

PR #102, opened 2017-11-03
Travis cleanup

zemanlx:
Big clean-up in `.travis.yml`. 
- Removed a lot of not used stuff.
- Get caching working again for OS X. 
- `stack default` was the same as `--resolver lts-9.10` therefore, I have replaced it with the latest LTS `--resolver lts`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,8 +65,8 @@ matrix:
       # Lint its own Dockerfile
       - docker run --rm -i hadolint:$(git describe --tags --dirty) < Dockerfile
       # Check that version in hadolint in the same as its `git describe`
-      # This can be enabled when there is an annotated tag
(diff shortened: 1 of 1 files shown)


lukasmartinelli:
> Big clean-up in .travis.yml.

So this will make builds a lot faster?

zemanlx:
It should definitely get better. Travis is limiting Mac builds so it still takes longer that I would like ... but it is free ;)

## Before
Sometimes Mac OS X cache was a bit bigger (to be able to upload in 180s) and that cause that all Mac builds in that branch were downloading/building everything again and again like this
image

## Now
Travis waits 300s until it timeouts.
image
Last long taking build is Docker images, but it also be better with

## New UPX for Docker image
image

lukasmartinelli:
Wow so much faster.

PR #128, opened 2017-12-07
Add stack install command to README for local install

samiralajmovic:
For those unfamiliar with Haskell and wondering what to do after `stack build` (includes me to a degree!).

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ build tool to build the binary.
 git clone https://github.com/hadolint/hadolint
 cd hadolint
 stack build
+stack install
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,6 +53,7 @@ build tool to build the binary.
 git clone https://github.com/hadolint/hadolint
 cd hadolint
 stack build
```
zemanlx: Thanks! Good idea @samiralajmovic. 

In this case, we can probably drop `stack build`.

zemanlx:
Thank you for contributing.

PR #134, opened 2017-12-12
Version pinning for alpine apk

### LINKED ISSUE #132: Support for Alpine's apk tool
- [ ] This is a bug report

### Expected behavior

Personally I prefer to stick to the much smaller alpine tags of any Docker image that supports it. The underlying [Alpine distro](https://alpinelinux.org/) uses it's own package manager, called [apk](https://wiki.alpinelinux.org/wiki/Alpine_Linux_package_management) which has the same features like apt-get, but is currently not linted.

### Actual behavior

It would be a good addition if hadolint was able to lint apk operations, too.  
Possible rules:
* Don't upgrade the system (via `apk upgrade` or `apk update && apk upgrade`)
* Use version pinning like apt-get (via `apk add asterisk=1.6.0.21-r0`)
* Delete cache (needs further investigation)

Lint Dockerfile with flaws in it, see no issues reported.

```bash
Haskell Dockerfile Linter UNKNOWN
```

This is weird. Just installed hadolint via Homebrew and got
```bash
==> Pouring hadolint-1.2.3.high_sierra.bottle.tar.gz
```

**Dockerfile**

```Dockerfile
FROM openjdk:8-jre-alpine

RUN apk update && apk upgrade && apk add \
  nodejs \
  nodejs-npm\
  curl
```

* macOS High Sierra 10.13.2 (17C88)
* Homebrew 1.4.0 (Homebrew/homebrew-core (git revision dbce; last commit 2017-12-11))

zemanlx:
### What I did

Added support for [alpine](https://alpinelinux.org/) `apk` package tool that should be similar to what we have for debian/ubuntu `apt-get`. As a bonus, I have added support for virtual packages so version pinning won't fail because of that.

I have added rules
- Do not use `apk upgrade`. https://github.com/hadolint/hadolint/wiki/DL3017
- Pin versions in apk add. Instead of `apk add <package>` use `apk add <package>=<version>`. https://github.com/hadolint/hadolint/wiki/DL3018
- Use the `--no-cache` switch to avoid the need to use `--update` and remove `/var/cache/apk/*` when done installing packages. https://github.com/hadolint/hadolint/wiki/DL3019

Closes #132. 

### How I did it

- Do not use `apk upgrade` - check if  `apk upgrade`  is part of args  of `RUN` instruction
- Pin versions in apk add. Instead of `apk add <package>` use `apk add <package>=<version>`. - check that command is `apk add`;  extract packages from command and then check if each package is pinned using `=`
- Use the `--no-cache` switch to avoid the need to use `--update` and remove `/var/cache/apk/*` when done installing packages. - check for `--no-cache` in `apk add` command

### How to verify it

I have added tests to `test/Spec.hs`

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,9 @@ Please [create an issue][] if you have an idea for a good rule.
 | [DL3014](https://github.com/hadolint/hadolint/wiki/DL3014)   | Use the `-y` switch.                                                                                                                                |
 | [DL3015](https://github.com/hadolint/hadolint/wiki/DL3015)   | Avoid additional packages by specifying --no-install-recommends.                                                                                    |
 | [DL3016](https://github.com/hadolint/hadolint/wiki/DL3016)   | Pin versions in `npm`.                                                                                                                              |
+| [DL3017](https://github.com/hadolint/hadolint/wiki/DL3017)   | Do not use `apk upgrade`.                                                                                                                           |
+| [DL3018](https://github.com/hadolint/hadolint/wiki/DL3018)   | Pin versions in apk add. Instead of `apk add <package>` use `apk add <package>=<version>`.                                                          |
+| [DL3019](https://github.com/hadolint/hadolint/wiki/DL3019)   | Use the `--no-cache` switch to avoid the need to use `--update` and remove `/var/cache/apk/*` when done installing packages.                        |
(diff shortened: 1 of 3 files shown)


zemanlx:
@lorenzo Can you please quick check this PR?

PR #145, opened 2017-12-19
Update stack resolver to lts-10.0

zemanlx:
### What I did

- Updated stack resolver to LTS-10.0 which contains:
  - language-docker 1.0.0
  - ShellCheck 0.4.7
- Added `package.yaml` which generates `hadolint.cabal` and enables us to use simpler format without redundancies
- Added alias in `.travis.yml` for fixed version of resolver, so next time we need to edit it only in place

All seems to work well. The only disadvantage I have found is that `ghc-mod` is not woking with `ghc-8.2.2`, https://github.com/DanielG/ghc-mod/issues/900

PS. `stack 1.6.1` is required to get it working.


zemanlx:
I am also thinking about putting `hadolint.cabal` into `.gitignore` as it is generated automatically with every `stack` command.

lorenzo:
Not having ghc-mod is one of the main reason I haven't upgraded my projects, but this looks good to me

zemanlx:
@lorenzo In that case, I can split updating LTS and use of `package.yaml` into two PR and this can wait until we have ghc-mod for 8.2.2

lorenzo:
I think there is not need to worry about it @zemanlx, let's just get this in.

zemanlx:
@lorenzo what do you think about adding cabal file to `.gitignore`?

lorenzo:
@zemanlx do you know of any other project doing that?

zemanlx:
@lorenzo I am pretty new in Haskell space, so short answer, no. That's why I am asking for your opinion :wink:

lorenzo:
@zemanlx I'll do some quick research, my view is that it should stay in version control, because there are tools that will require it there. Will get back to you after I shop around

lorenzo:
@zemanlx Add it to gitignore. See projects do the same, and I can't personally think of a reason to keep it around

Issue #146, opened 2018-01-03
[Bug] Break on comments inside dockerfile


Yivan:
Hello

Thanks for this great tools and sharing it with us.

It seems to have problem with comment inside multiline apt-get.
If you try to validate this dockerfile (official mariadb):
https://raw.githubusercontent.com/docker-library/mariadb/506cd7c6b9d6c53d7b514a2c94db5d3daee832fd/10.3/Dockerfile

You will see hadolint cannot parse it and breaks.
Maybe comments should be automaticaly stripped out before analyse ? (removing the comment lines)

Thanks

zemanlx:
Thank you @Yivan for reporting, we know that it is an issue. @lorenzo has done a great progress in our parser [language-docker](https://github.com/hadolint/language-docker) which is used in _hadolint_ and did a release today. I believe this issue disappear as soon as we use `language-docker-2.0.0`

@lorenzo do you plan to make a PR for that?

lorenzo:
@zemanlx I tagged 2.0 today, will try sending a PR to this repo with the upgrade soon

Yivan:
Thank you for you fast fix! All work nice now.

Issue #76, opened 2017-02-01
Linter doesn't handle comments within the `RUN` shell block


dnwe:
Having comments with a `RUN` block are perfectly valid shell, and `docker build` handles them without issue.

e.g., (contrived example)
```
$ cat Dockerfile
FROM alpine:edge

RUN apk --no-cache add zsh \
  # now add vim
  && apk --no-cache add vim
```

```
% docker build -t temp . 
Sending build context to Docker daemon 2.048 kB
Step 1/2 : FROM alpine:edge
 ---> 3fc33d6d5e74
Step 2/2 : RUN apk --no-cache add zsh   && apk --no-cache add vim
 ---> Running in 842e35506e47
fetch http://dl-cdn.alpinelinux.org/alpine/edge/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/edge/community/x86_64/APKINDEX.tar.gz
(1/4) Installing ncurses-terminfo-base (6.0-r7)
(2/4) Installing ncurses-terminfo (6.0-r7)
(3/4) Installing ncurses-libs (6.0-r7)
(4/4) Installing zsh (5.3.1-r0)
Executing zsh-5.3.1-r0.post-install
Executing busybox-1.24.2-r9.trigger
OK: 17 MiB in 15 packages
fetch http://dl-cdn.alpinelinux.org/alpine/edge/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/edge/community/x86_64/APKINDEX.tar.gz
(1/2) Installing lua5.2-libs (5.2.4-r2)
(2/2) Installing vim (8.0.0187-r0)
Executing busybox-1.24.2-r9.trigger
OK: 43 MiB in 17 packages
 ---> 760d383ada47
Removing intermediate container 842e35506e47
Successfully built 760d383ada47
```

However, hadolint currently fails to handle them:
```
% docker run --rm -i lukasmartinelli/hadolint <Dockerfile
/dev/stdin:5:1 unexpected '&' expecting space, "\t", "ONBUILD", "FROM", "COPY", "RUN", "WORKDIR",
"ENTRYPOINT", "VOLUME", "EXPOSE", "ENV", "ARG", "USER", "LABEL", "STOPSIGNAL", "CMD", "SHELL",
"MAINTAINER", "ADD", "#", "HEALTHCHECK" or end of input
```

Unless you strip them out with sed:
```
[...]

Issue #53, opened 2016-08-22
EXPOSE with variables doesn't work


ttres:
The Dockerfile validation fails when the ports arguments to the EXPOSE command are declared as environment variables. Take this dummy, useless code as example:

``` Docker
FROM alpine:latest
ENV PORT=8080
EXPOSE $PORT
```

This is the output:

``` Bash
$ docker run --rm -i lukasmartinelli/hadolint < Dockerfile
"/dev/stdin" (line 3, column 8):
unexpected '$'
expecting space, "\t", "ONBUILD", "FROM", "COPY", "RUN", "WORKDIR", "ENTRYPOINT", "VOLUME", "EXPOSE", "ENV", "ARG", "USER", "LABEL", "STOPSIGNAL", "CMD", "MAINTAINER", "ADD", "#" or end of input
```

But the image inspection show that the correct port was exposed:

``` JSON
$ docker build --tag=sample -f Dockerfile .
$ docker inspect sample | grep -A2 Exposed
            "ExposedPorts": {
                "8080/tcp": {}
            },
```

Declaring variables with the exposed ports allow for reuse inside the Dockerfile.

lukasmartinelli:
> The Dockerfile validation fails when the ports arguments to the EXPOSE command are declared as environment variables. Take this dummy, useless code as example:

I don't support variables anywhere right now.
Perhaps I could just ad another preprocessing pipeline that replaces the env vars with the declared values.

1. Scan through file once
2. Build AST
3. Filter out ENV statements
4. Replace `$ENV` in file with `$ACTUAL_VALUE`
..
1. And then start parsing for the real thing (scan through file, build ast, do linting)

In a language parser this stuff is probably bad due to performance issues but here it is not so critical.

ttres:
> In a language parser this stuff is probably bad due to performance issues but here it is not so critical.

I understand the issue.

Your approach sounds good to me.

StevenACoffman:
Is there a pull request where this work is being done?

Issue #38, opened 2016-07-14
Exposed udp ports don't work


schoentoon:
When I have a Dockerfile that is exposing udp ports using the `EXPOSE 53/udp` syntax I get the following error.
`"<string>" (line 18, column 12):
unexpected '/'
expecting space, "\t", "ONBUILD", "FROM", "COPY", "RUN", "WORKDIR", "ENTRYPOINT", "VOLUME", "EXPOSE", "ENV", "ARG", "USER", "LABEL", "STOPSIGNAL", "CMD", "MAINTAINER", "ADD", "#" or end of input`

PR #147, opened 2018-01-03
Upgrade language-docker to version 2.0

lorenzo:
This meant that of the Check rules had to be removed, as they now are parse errors

Fixes https://github.com/hadolint/hadolint/issues/146
Fixes https://github.com/hadolint/hadolint/issues/76
Fixes https://github.com/hadolint/hadolint/issues/53
Fixes https://github.com/hadolint/hadolint/issues/38

### DIFF
--- a/src/Hadolint/Rules.hs
+++ b/src/Hadolint/Rules.hs
@@ -226,8 +226,7 @@ noLatestTag = instructionRule code severity message check
     message =
         "Using latest is prone to errors if the image will ever update. Pin the version explicitly \
         \to a release tag."
-    check (From (TaggedImage _ tag _)) =
-        not (isPrefixOf "latest AS" tag || isPrefixOf "latest as" tag || tag == "latest")
+    check (From (TaggedImage _ tag _)) = tag /= "latest"
     check _ = True
 
 aptGetVersionPinned = instructionRule code severity message check
@@ -465,7 +464,8 @@ copyInsteadAdd = instructionRule code severity message check
     code = "DL3020"
     severity = ErrorC
     message = "Use COPY instead of ADD for files and folders"
-    check (Add (AddArgs srcs _ _)) = and [isArchive src || isUrl src | SourcePath src <- toList srcs]
+    check (Add (AddArgs srcs _ _)) =
+        and [isArchive src || isUrl src | SourcePath src <- toList srcs]
     check _ = True
 
 useShell = instructionRule code severity message check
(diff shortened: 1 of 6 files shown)


src/Hadolint/Rules.hs
```diff
@@ -311,33 +311,23 @@ useAdd = instructionRule code severity message check
     code = "DL3010"
     severity = InfoC
     message = "Use ADD for extracting archives into an image"
-    check (Copy src dst) = and [not (format `isSuffixOf` src) | format <- archive_formats]
-    check _ = True
-    archive_formats = [".tar", ".gz", ".bz2", "xz"]
-
-exposeMissingArgs = instructionRule code severity message check
```
lorenzo: this is now caught as a parse error

src/Hadolint/Rules.hs
```diff
@@ -311,33 +311,23 @@ useAdd = instructionRule code severity message check
...
-  where
-    code = "DL3021"
-    severity = ErrorC
-    message = "EXPOSE requires at least one argument"
-    check (Expose (Ports ports)) = not (null ports)
-    check (Expose (PortStr "")) = False
-    check _ = True
-
-copyMissingArgs = instructionRule code severity message check
```
lorenzo: This also became a parse error

src/Hadolint/Rules.hs
```diff
@@ -226,7 +226,7 @@ noLatestTag = instructionRule code severity message check
     message =
         "Using latest is prone to errors if the image will ever update. Pin the version explicitly \
         \to a release tag."
-    check (From (TaggedImage _ tag)) =
+    check (From (TaggedImage _ tag _)) =
```
zemanlx: I believe that tag and name are now nicely separated, we can replace this _hack_ by 

```haskell
check (From (TaggedImage _ tag _)) = tag /= "latest"
```

src/Hadolint/Rules.hs
zemanlx: Can you please also remove it from `README.md`?

src/Hadolint/Rules.hs
zemanlx: Can you please also remove it from `README.md`?

lorenzo: It was not in the readme!

src/Hadolint/Rules.hs
lorenzo: ah, good catch

lukasmartinelli:
giphy 80

src/Hadolint/Rules.hs
zemanlx: Sorry, my bad. I was expecting that all checks are in README.

lorenzo:
I'm going to release a new version

zemanlx:
@lorenzo cool, please follow https://github.com/hadolint/hadolint/blob/master/docs/RELEASE.md

PR #149, opened 2018-01-06
Added support for a config file

### LINKED ISSUE #118: Use config file to ignore specific rules
Would be nice to have config file like `.hadolintrc` to disable specific rules check
Most linters have this feature

lorenzo:
CLI options take precedence over the config file
Closes #118

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -79,7 +79,7 @@ applyConfig o = do
         Just config -> parseAndApply config
   where
     findConfig = do
-        localConfigFile <- (</> "hadolint.yaml") <$> getCurrentDirectory
+        localConfigFile <- (</> ".hadolint.yaml") <$> getCurrentDirectory
         configFile <- getXdgDirectory XdgConfig "hadolint.yaml"
         listToMaybe <$> filterM doesFileExist [localConfigFile, configFile]
     parseAndApply config = do
@@ -98,7 +98,7 @@ applyConfig o = do
             _ ->
                 error $
                 "Error parsing your config file in  '" ++
-                config ++ "': " ++ (Yaml.prettyPrintParseException err)
+                config ++ "': " ++ Yaml.prettyPrintParseException err
 
 -- | Support UNIX convention of passing "-" instead of "/dev/stdin"
 parseFilename :: String -> String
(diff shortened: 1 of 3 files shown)


app/Main.hs
```diff
@@ -38,19 +53,53 @@ printChecks checks = do
...
+applyConfig o@(LintOptions _ (_:_) _) = return o
+applyConfig o = do
+    theConfig <- findConfig
+    case theConfig of
+        Nothing -> return o
+        Just config -> parseAndApply config
+  where
+    findConfig = do
+        localConfigFile <- (</> "hadolint.yaml") <$> getCurrentDirectory
+        configFile <- getXdgDirectory XdgConfig "hadolint.yaml"
```
zemanlx: Would you mind to have it hidden by defaul, eg `.hadolint.yaml`?

zemanlx:
I have also asked appveyor to increase a limit for our project so I hope it stops failing soon.

app/Main.hs
lorenzo: Support both hidden and not hidden?

zemanlx: I think hidden(starting with `.`) like `.travis.yml` or `.hindent.yaml` or `.zshrc` is enough.

zemanlx:
When you will be editing README, can you please also change https://github.com/hadolint/hadolint#ast appropriately, when `Syntax.hs` is no longer part of hadolint?

lorenzo:
@zemanlx done

lorenzo:
Only the local one is take in account if both exist

lorenzo:
@zemanlx any answers from appveyor?

zemanlx:
I have just got a response a few minutes ago...rebuilding now. I have noticed that binaries are now much bigger. 

[16MB](https://ci.appveyor.com/project/hadolint/hadolint/build/1.0.75/artifacts) vs [26MB](https://ci.appveyor.com/project/hadolint/hadolint/build/1.0.81/artifacts) 

Is it possible to avoid this? Not a big issue, just asking :wink:

lorenzo:
@zemanlx it is likely not an issue once the final release applies `upx` on it, but I don't think we can reduce the size of the binary without compressing it with upx first

lorenzo:
I'll merge and tag a new release

PR #162, opened 2018-02-02
Improve dl3006

### LINKED ISSUE #161: DL3006 rule fails in multi-stage build FROM
- [x] This is a bug report

### Expected behavior
The rule should not fail when `FROM` base image uses an other stage in a multi-stage build Dockerfile

### Actual behavior
The rule fails

The second build stage `unit-test` of the following Dockerfile is using the previous stage as it's base image which is totally fine but the rule require that we specify a tag and no tag exists for a stage.

The Dockerfile is an example and the problem with the rule is not related to the language at all.

Using `hadolint/hadolint:latest` image to lint a multi-stage Dockerfile

```Dockerfile
FROM golang:1.9.3-alpine3.7 AS build

... (building my go app here)

FROM build as unit-test

... (testing my go app here)

FROM alpine:3.7
...
```

lorenzo:
Fixes #161

### DIFF
--- a/src/Hadolint/Rules.hs
+++ b/src/Hadolint/Rules.hs
@@ -211,14 +211,20 @@ noAptGetUpgrade = instructionRule code severity message check
     check (Run args) = not $ isInfixOf ["apt-get", "upgrade"] args
     check _ = True
 
-noUntagged = instructionRule code severity message check
+noUntagged dockerfile = instructionRule code severity message check dockerfile
   where
     code = "DL3006"
     severity = WarningC
     message = "Always tag the version of an image explicitly."
-    check (From (UntaggedImage image _)) = image == "scratch"
+    check (From (UntaggedImage image _)) = image == "scratch" || image `elem` aliasedImages
     check (From TaggedImage {}) = True
     check _ = True
+    aliasedImages =
+        unImageAlias <$>
+        [a | From (UntaggedImage _ (Just a)) <- instr] ++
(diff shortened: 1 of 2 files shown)


package.yaml
```diff
@@ -1,5 +1,5 @@
 name: hadolint
-version: '1.2.6'
+version: '1.3.0'
```
lorenzo: Forgot to push this to master

src/Hadolint/Rules.hs
```diff
@@ -211,14 +211,20 @@ noAptGetUpgrade = instructionRule code severity message check
...
     code = "DL3006"
     severity = WarningC
     message = "Always tag the version of an image explicitly."
-    check (From (UntaggedImage image _)) = image == "scratch"
+    check (From (UntaggedImage image _)) = image == "scratch" || image `elem` aliasedImages
     check (From TaggedImage {}) = True
     check _ = True
+    aliasedImages =
+        unImageAlias <$>
+        [a | From (UntaggedImage _ (Just a)) <- instr] ++
```
lukasmartinelli: I don't quite understand what this does anymore :( My haskell is decaying.

lorenzo: It gets all elements of the instr list that can be pattern matched to that expression

lorenzo: and returns the variable a, as the elements of the resulting list

zemanlx:
Nice work. Thanks. I will probably add check for lint and version in Travis later.

PR #163, opened 2018-02-04
Added new rules to check FROM aliases

lorenzo:
Added 3 new rules:

*  `COPY --from` should reference a previously defined `FROM` alias  
* `COPY --from` cannot reference its own `FROM` alias
* `FROM` aliases (stage names) must be unique

The 3 of them check for error conditions in the dockerfile, they are not just best practices.

Additionally, re-factored the traversal methods like `instructionRule` to include `instructionRuleLine` and `instructionRuleState`. the new functions allow more complex rule checking.

In particular `instructionRuleState` allows the check function to accumulate state in an a variable that can be used to compare the currently being inspected instruction to some previous checks.


lorenzo:
I'll release a new version

PR #166, opened 2018-02-24
Linting and version check

zemanlx:
### What I did

New build in Travis will check for Haskell (hlint) and Markdown (remark) linting issues. I have also added a check if a version in `package.yaml` is equal or higher than a version in `git describe` to avoid situation when we forgot to update a version number in `package.yaml` before release. This, of course, won't stop us to do it, but at least alerts us when we forgot.

Out of scope changes:
- removed docker update - no longer needed for a multistage build. Travis is now using `17.09-ce` by default. It saves us some time.
- update to LTS-10.7 and removed optparse-applicative restriction. I have never seen it failing because of any new version of that package. 

### How to verify it

Travis baby.

### DIFF
--- a/.remarkrc.yaml
+++ b/.remarkrc.yaml
@@ -1,0 +1,9 @@
+plugins:
+  preset-lint-consistent:
+  preset-lint-markdown-style-guide:
+  preset-lint-recommended:
+  validate-links:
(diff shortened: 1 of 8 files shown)


lorenzo:
This looks great!

PR #175, opened 2018-03-18
UTF input

### LINKED ISSUE #172: Fails with dockerfile containing umlauts (special character)
- [x] This is a bug report

### Expected behavior

Run as usual and output results.

### Actual behavior

Having umlauts and possilbly other special characters in the Dockerfile breaks hadolint. hadolint: /dev/stdin: hGetContents: invalid argument (invalid byte sequence)

Minimal example: output is
```
docker run --rm -i hadolint/hadolint < Dockerfile.test2
hadolint: /dev/stdin: hGetContents: invalid argument (invalid byte sequence)
```

Dockerfile.test2
```Dockerfile
FROM debian:latest

LABEL maintainer="ö"

RUN echo "Hallo"
```

Running without the `ö` does work as espected. It outputs:
```
/dev/stdin:1 DL3007 Using latest is prone to errors if the image will ever update. Pin the version explicitly to a release tag
```

```bash
Haskell Dockerfile Linter v1.5.0-13-gc472436
```

zemanlx:
### What I did

Added Debian image (like our usual BusyBox images but with `-debian` suffix) that uses `locale` so that we can setup `LC_ALL` to UTF-fy locale and therefore handle UTF-8 characters in a `Dockerfile`.

Fixes #172.

### How I did it

- Added new instructions in our `Dockerfile` for Debian
- Added hooks for Docker Hub to build also Debian image

PS Now we have two `CMD` instructions in `Dockerfile` so I have added DL4003 to ignore during Travis/Appveyor checks. 

PS2 Docker Hub build is now really long. I will probably investigate if we can speed up build time later.

### How to verify it

Try example from #172 using latest Debian image from https://hub.docker.com/r/zemanlx/hadolint/tags/


lorenzo:
@zemanlx Can't we just offer the debian image and not the other? I don't see a lot of benefit in offering an image that does not work correctly for all cases

zemanlx:
@lorenzo I would like to keep continuity in what we have. It is not unusual to have more distro-images of one product commit. Also, I would say that most people do not need non-asci characters in Dockerfiles. For me as a person who is using it in CI, every added second of build/test is important. Our current tiny image may be important for people that have a slower internet connection. Plus having two images do not cost us anything.

lorenzo:
@zemanlx the cost is in confusion for the users. How much bigger is the debian based image?

jbergstroem:
> @zemanlx said:
> Also, I would say that most people do not need non-asci characters in Dockerfiles.

Most editors default to UTF-8. Combine that with for instance using your name via `LABEL` and boom.

zemanlx:
OK, I will rebase this and remove _BusyBox_ and replace it by _Debian_.

jbergstroem:
> @zemanlx said:
> Our current tiny image may be important for people that have a slower internet connection.

I think in container-land one should always optimize for size. There is literally no benefit to a larger image since the interface is the application, not its environment.

zemanlx:
@jbergstroem AFAIK Alpine is not supported by `GHC`.

zemanlx:
I will probably keep this branch (and close this PR) and create a new one for replacement of BusyBox

zemanlx:
Superseeded by https://github.com/hadolint/hadolint/pull/178.

PR #178, opened 2018-03-20
Switch default distro image to Debian

### LINKED ISSUE #172: Fails with dockerfile containing umlauts (special character)

zemanlx:
### What I did

Replaced BusyBox by Debian image that uses `locale` so that we can setup `LC_ALL` to UTF-fy locale and therefore handle UTF-8 characters in a `Dockerfile`.

Fixes #172.

### How to verify it

Try example from #172 using latest Debian image from https://hub.docker.com/r/zemanlx/hadolint/tags/


lorenzo:
Great!

PR #183, opened 2018-03-25
Fixed the way pip and npm packages are found in RUN instructions

### LINKED ISSUE #181: incorrectly flagging `python -m pip install ...`
- [x] This is a bug report

### Expected behavior

hadolint should not flag an error for a pinned pip package when run like `python -m pip install ...`.

### Actual behavior

I get an error when using `python -m pip install`, see below.

```bash
% docker run --rm -i hadolint/hadolint hadolint --version
Haskell Dockerfile Linter v1.6.0-0-gbd99295
% docker images | grep hadolint
hadolint/hadolint                                      latest              2b2411ca71ff        14 hours ago        59MB
```

**Dockerfile**

```Dockerfile
RUN pip install git+https://github.com/example/example@123abcdef
RUN pip install example==1.2.3
RUN python -m pip install example==1.2.3
RUN python -m pip install git+https://github.com/example/example@123abcdef
```

**Output**

```bash
% docker run --rm -i hadolint/hadolint < Dockertest
/dev/stdin:3 DL3013 Pin versions in pip. Instead of `pip install <package>` use `pip install <package>==<version>`
/dev/stdin:4 DL3013 Pin versions in pip. Instead of `pip install <package>` use `pip install <package>==<version>`
```

### See Also

Issue #67 is similar, and something like PR #99 may fix this. Adding `"python", "-m"` to the same location in [Rules.hs](https://github.com/hadolint/hadolint/pull/99/files#diff-7086fb4793c389a29a1ea9410da4098eR269) might fix the test case above, but note that in the original Dockerfile where I found this I'm using `RUN python3.6 -m pip install ...` and it doesn't seem ideal to have to enumerate all possible python versions in that list. I'm not really a haskell hacker so there may be a better solution.

lorenzo:
fixes #181 

### What I did

Re-implemented how `pip install` and `npm install` packages are found, by not hardcoding as much as before.

### How to verify it

Added a new test case

### DIFF
--- a/src/Hadolint/Rules.hs
+++ b/src/Hadolint/Rules.hs
@@ -484,7 +484,7 @@ pipVersionPinned = instructionRule code severity message check
 findPackages :: [String] -> [String]
 findPackages = takeWhile (not . isEnd) . dropWhile isFlag
   where
-    isEnd word = isFlag word || word == "&&"
+    isEnd word = isFlag word || word `elem` ["&&", "||", ";"]
     isFlag ('-':_) = True
     isFlag _ = False
 
--- a/src/Hadolint/Rules.hs
+++ b/src/Hadolint/Rules.hs
@@ -484,7 +484,7 @@ pipVersionPinned = instructionRule code severity message check
 findPackages :: [String] -> [String]
 findPackages = takeWhile (not . isEnd) . dropWhile isFlag
   where
-    isEnd word = isFlag word || word `elem` ["&&", "||", ";"]
+    isEnd word = isFlag word || word `elem` ["&&", "||", ";", "|"]
     isFlag ('-':_) = True
     isFlag _ = False
 
--- a/src/Hadolint/Rules.hs
+++ b/src/Hadolint/Rules.hs
@@ -3,7 +3,8 @@
 module Hadolint.Rules where
 
 import Control.Arrow ((&&&))
-import Data.List (isInfixOf, isPrefixOf, isSuffixOf, mapAccumL)
+import Data.List
+       (dropWhile, isInfixOf, isPrefixOf, isSuffixOf, mapAccumL)
 import Data.List.NonEmpty (toList)
 import Data.List.Split (splitOneOf)
 import Hadolint.Bash
@@ -377,7 +378,8 @@ aptGetCleanup dockerfile = instructionRuleState code severity message check Noth
     imageIsUsedLater line baseimage =
         case fromAlias baseimage of
             Nothing -> True
-            Just (ImageAlias alias) -> alias `elem` [i | (l, i) <- allImageNames dockerfile, l > line]
+            Just (ImageAlias alias) ->
+                alias `elem` [i | (l, i) <- allImageNames dockerfile, l > line]
 
 dropOptionsWithArg :: [String] -> [String] -> [String]
 dropOptionsWithArg os [] = []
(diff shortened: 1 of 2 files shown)


src/Hadolint/Rules.hs
```diff
@@ -506,24 +520,18 @@ npmVersionPinned = instructionRule code severity message check
         \<package>@<version>`"
     check (Run (Arguments args)) = all versionFixed (packages args)
     check _ = True
-    packages :: [String] -> [String]
```
zemanlx: out of curiosity.. why do you remove type signatures?

src/Hadolint/Rules.hs
```diff
@@ -459,33 +461,45 @@ pipVersionPinned = instructionRule code severity message check
...
-        noOption arg = arg `notElem` commandName && not (isFlag arg)
-        commandName = ["pip", "pip2", "pip3", "install"]
-        isFlag ('-':_) = True
-        isFlag _ = False
+
+-- | Returns all the packages after pip install
+findPackages :: [String] -> [String]
+findPackages = takeWhile (not . isEnd) . dropWhile isFlag
+  where
+    isEnd word = isFlag word || word == "&&"
```
zemanlx: I suppose `&&` is to stop before next chained command. If that is the case, we should probably also count on others like `;` and `||` or `|`. I believe there will be a use case for them.

src/Hadolint/Rules.hs
lorenzo: They were very noisy in this context, they were making the reading of the code quite difficult for me. I can bring them back if you find the useful.

src/Hadolint/Rules.hs
lorenzo: Yeah, good idea. I'll add `;` and `||`

src/Hadolint/Rules.hs
zemanlx: OK, no problem

src/Hadolint/Rules.hs
```diff
@@ -484,7 +484,7 @@ pipVersionPinned = instructionRule code severity message check
 findPackages :: [String] -> [String]
 findPackages = takeWhile (not . isEnd) . dropWhile isFlag
   where
-    isEnd word = isFlag word || word == "&&"
+    isEnd word = isFlag word || word `elem` ["&&", "||", ";"]
```
zemanlx: What about pipe `|` ?

lorenzo: out ops, forgot about it.

Aside: do you actually use a single pipe in bash inside docker? I've never written one, let alone in a dockerfile?

zemanlx: I have it usually in this scenario

```Dockerfile
RUN curl -sSL https://get.docker.com/builds/Linux/x86_64/docker-${DOCKER_VERSION}.tgz \
  | tar -xzv --strip-components=1 -C /bin \
  && chmod +x /bin/docker-* \
  && ln -s /bin/docker /bin/docker-${DOCKER_VERSION}
```

lorenzo: ok, but not in a `pip install` right? I don't think it will hurt leaving it there, so no worries

zemanlx: Yes, but I am not a heavy user of python projects. I believe there is somebody who will find it useful :smile:

PR #189, opened 2018-04-15
Permit squishing of single-letter options to apt-get install

### LINKED ISSUE #188: DL3008 false positive caused by squishing apt-get -y and -qq options into -yqq
- [x] This is a bug report

### Expected behavior
hadolint should not report DL3008 when Dockerfile RUN directive contains apt-get install command that squishes the -y and -qq options into -yqq (or squishes any other single-letter options that take no arguments).

### Actual behavior
hadolint DOES report DL3008 in this case.

```bash
Haskell Dockerfile Linter 1.4.0-no-git
```

**Dockerfile** (if relevant)

```Dockerfile
FROM debian:jessie-slim
RUN apt-get update -qq \
    && apt-get install --no-install-recommends -yqq wget=1.16-1+deb8u4 \
    && rm -rf /var/lib/apt/lists/*
```

mzraly:
Fixes #188

### What I did
Modified aptGetPackages to treat any word in the command that starts with a "-" as an option.

### How to verify it
Existing test case modified to include squished option "-yqq".

### DIFF
--- a/src/Hadolint/Rules.hs
+++ b/src/Hadolint/Rules.hs
@@ -349,8 +349,8 @@ aptGetPackages args =
         , isAptGetInstall cmd
         ]
   where
-    noOption arg = arg `notElem` options && not ("--" `isPrefixOf` arg)
-    options = ["apt-get", "install", "-d", "-f", "-m", "-q", "-y", "-qq"]
+    noOption arg = arg `notElem` words && not ("-" `isPrefixOf` arg)
+    words = ["apt-get", "install"]
 
 aptGetCleanup dockerfile = instructionRuleState code severity message check Nothing dockerfile
   where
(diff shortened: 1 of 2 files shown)


src/Hadolint/Rules.hs
```diff
@@ -349,8 +349,8 @@ aptGetPackages args =
         , isAptGetInstall cmd
         ]
   where
-    noOption arg = arg `notElem` options && not ("--" `isPrefixOf` arg)
-    options = ["apt-get", "install", "-d", "-f", "-m", "-q", "-y", "-qq"]
+    noOption arg = arg `notElem` words && not ("-" `isPrefixOf` arg)
+    words = ["apt-get", "install"]
```
lorenzo: Would you mind naming this something else? words is a very commonly known function.

You can in-line the list, if coming up with a name is difficult.

lorenzo:
Thanks!

PR #206, opened 2018-06-06
Update ShellCheck to 0.5.0

zemanlx:
### What I did

A new version of ShellCheck has been released last week so update to keep us :sunglasses:

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -33,7 +33,7 @@ library:
   dependencies:
     - &bytestring bytestring >=0.10
     - &split split >=0.2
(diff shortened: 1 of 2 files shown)


lorenzo:
cool!

lorenzo:
@zemanlx there are new rules added, do you want to add them to the readme? https://github.com/koalaman/shellcheck/blob/master/CHANGELOG.md#added

zemanlx:
@lorenzo I think we are not listing all of them.

PR #210, opened 2018-06-07
Building debian slim image

mrferos:
### What I did
Builds a debian slim image based off the main hadolint image, the debian slim images are named `hadolint/hadolint:debian-slim-<version>`
### How I did it
The debian slim Dockerfile uses the hadolint image as the base image and then copies the binary from it in the main stage that uses the debian slim image
### How to verify it
After build:
```bash
docker run --rm -i hadolint/hadolint:debian-slim-latest < Dockerfile
```

### DIFF
--- a/docker/Dockerfile-debian-slim
+++ b/docker/Dockerfile-debian-slim
@@ -1,0 +1,10 @@
+ARG image_name
+FROM $image_name as builder
+FROM debian:stretch-slim
+
+RUN apt-get update \
+ && apt-get install --no-install-recommends -y git=1:2.11.*
+
+COPY --from=builder /bin/hadolint /bin/hadolint
+
+CMD ["/bin/hadolint", "-"]
--- a/docker/hooks/build
+++ b/docker/hooks/build
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -x
 
 echo "=> Building main container"
 docker build \
@@ -7,6 +7,6 @@ docker build \
 
 echo "=> Building debian container"
 docker build \
-    --build-arg image_name $IMAGE_NAME \
-    --tag ${IMAGE_NAME/:/:debian-slim}
+    --build-arg image_name=$IMAGE_NAME \
+    --tag ${IMAGE_NAME/:/:debian-slim-} \
     --file Dockerfile-debian-slim ..
--- a/docker/hooks/build
+++ b/docker/hooks/build
@@ -1,6 +1,12 @@
 #!/bin/bash
 
-echo "=> Building the container"
+echo "=> Building main container"
 docker build \
   --tag $IMAGE_NAME \
   --file Dockerfile ..
+
+echo "=> Building debian container"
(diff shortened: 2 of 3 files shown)


lorenzo:
@zemanlx can you verify this is correct?

docker/Dockerfile-debian-slim
```diff
@@ -0,0 +1,10 @@
+ARG image_name
+FROM $image_name as builder
+FROM debian:stretch-slim
+
+RUN apt-get update \
+ && apt-get install --no-install-recommends -y git=1:2.11.*
```
zemanlx: Why is git needed?

docker/Dockerfile-debian-slim
```diff
@@ -0,0 +1,10 @@
+ARG image_name
```
zemanlx: I would prefer to keep everything in one Dockerfile

docker/hooks/build
```diff
@@ -1,6 +1,12 @@
...
+echo "=> Building main container"
 docker build \
   --tag $IMAGE_NAME \
   --file Dockerfile ..
+
+echo "=> Building debian container"
+docker build \
+    --build-arg image_name $IMAGE_NAME \
+    --tag ${IMAGE_NAME/:/:debian-slim}
+    --file Dockerfile-debian-slim ..
```
zemanlx: Please add newlines.

docker/hooks/build
```diff
@@ -7,6 +7,6 @@ docker build \
 
 echo "=> Building debian container"
 docker build \
-    --build-arg image_name $IMAGE_NAME \
-    --tag ${IMAGE_NAME/:/:debian-slim}
+    --build-arg image_name=$IMAGE_NAME \
+    --tag ${IMAGE_NAME/:/:debian-slim-} \
```
zemanlx: I would also prefer `stretch` or `slim` rather than `debian-slim`

zemanlx:
Plus we need both of them build in Travis an then tested.

docs/INTEGRATION.md
```diff
@@ -37,6 +37,15 @@ lint_dockerfile:
     - docker run --rm -i hadolint/hadolint < Dockerfile
 ```
 
+Alternatively if you would rather run the hadolint image directly, there is a `debian-slim` variant available:
+```yaml
+lint_dockerfile:
+  stage: lint
+  image: hadolint/debian-slim-<version>
```
mrueg: should this be hadolint/hadolint:debian-slim-<version> ?

mrferos:
Hi everyone, sorry about the lack of activity, after talking about it with the developers, we're fine with just taking the route describe in the documentation for gitlab-ci. We don't run any other instructions on the docker-line steps so we don't gain much from enabling the behavior this PR would give us.

docs/INTEGRATION.md
lorenzo: Yes

PR #211, opened 2018-06-07
Set `LC_ALL=` for Travis

### LINKED ISSUE #173: Crash on Linux with specific locale settings
- [x] This is a bug report

When running with the following locale settings:
```
LANG=en_US.UTF-8
LANGUAGE=en_US:en
LC_CTYPE=en_US.UTF-8
LC_NUMERIC=C
LC_TIME=en_US.UTF-8
LC_COLLATE=C
LC_MONETARY=C
LC_MESSAGES=en_US.UTF-8
LC_PAPER=de_DE.UTF-8
LC_NAME="en_US.UTF-8"
LC_ADDRESS="en_US.UTF-8"
LC_TELEPHONE="en_US.UTF-8"
LC_MEASUREMENT="en_US.UTF-8"
LC_IDENTIFICATION="en_US.UTF-8"
LC_ALL=
```
hadolint crashes immediately with the error message:
```
 hadolint: loadlocale.c:129: _nl_intern_locale_data: Assertion `cnt < (sizeof (_nl_value_type_LC_TIME) / sizeof (_nl_value_type_LC_TIME[0]))' failed.
```
If `LC_ALL` is set to `C` before, hadolint can be used as usual.
I use the latest version 1.5.0 on Debian buster. On macOS, hadolint works fine without setting `LC_ALL`.

zemanlx:
### What I did

Fixes #173 

When `LC_ALL` is set during compilation it causes that binary is crashing as described in #173 (in case `LC_ALL` is not set on end machine).

> Using LC_ALL is strongly discouraged as it overrides everything. Please use it only when testing and never set it in a startup file.
https://wiki.debian.org/Locale

### How I did it

Tested release binaries 1.5.0 and 1.7.0 in Debian Buster with locale as described in #173 and they were crashing. Then I mounted my locally build binary with `LC_ALL` set to "" and that one works :tada:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,9 +17,9 @@ matrix:
   include:
     # The Stack builds. We can pass in arbitrary Stack arguments via the ARGS
     # variable, such as using --stack-yaml to point to a different file.
-    - env: ARGS="--resolver lts"
+    - env: &lts-latest ARGS="--resolver lts" LC_ALL=
 
-    - env: &lts-fixed ARGS="--resolver lts-11.12"
+    - env: &lts-fixed ARGS="--resolver lts-11.12" LC_ALL=
       addons:
         artifacts: {paths: [./releases]}
       before_deploy:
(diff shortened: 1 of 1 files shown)


zemanlx:
@lorenzo Looks like that there is more/different changes to make it effective in Travis :disappointed: 

PS: please do not merge my PRs, only review and I merge them when I am ready to deal with consequences :wink:

lorenzo:
@zemanlx haha, deal. What other changes do we need in travis to make this work?

zemanlx:
looks like that they set LC_ALL back to E=en_US.UTF-8, in the middle somewhere. I am trying to run debug builds now :scream:

PR #220, opened 2018-06-12
Split Hadolint Dockerfile into Scratch and Debian-Slim

### LINKED ISSUE #218: Please readd /bin/sh to hadolint docker image
- [x] This is a bug report

### Expected behavior
/bin/sh available on hadolint docker image
### Actual behavior
None found
Since this change, there's no /bin/sh in the hadolint docker image:
https://github.com/hadolint/hadolint/commit/8c2642232cbc4daa27f553bc3b4f1ddb63ba4b70

I added hadolint to my CI system and which executes steps within a shell, that make /bin/sh necessary. Before building my own images, I'd like to ask if you could you consider readding /bin/sh and maybe publish the other one as hadolint-minimal?

wevcode:
This PR splits hadolint's Dockerfile into two seperate Dockerfiles to build two different docker images.
Should fix #218 and is a follow-up of #210.


wevcode:
@lorenzo How do you build the finale Dockerimage which will be pushed to Docker Hub? Where do the environment variables are set?

docker/hooks/build
```diff
@@ -1,6 +1,11 @@
 #!/bin/bash
 
-echo "=> Building the container"
+echo "=> Building container: 'scratch' (default)"
 docker build \
-  --tag $IMAGE_NAME \
-  --file Dockerfile ..
+  --tag $IMAGE_NAME-scratch \
```
lorenzo: I don't think this should change

lorenzo:
@wevcode the variable is set by the automatic builder in hub.docker.com  

Can you address the pending comments in #210 ?

lorenzo:
@wevcode the variables are documented here. Note that `$IMAGE_NAME` contains both the name and tag

lorenzo:
This file would also need to be changed https://github.com/hadolint/hadolint/blob/master/docker/hooks/push

wevcode:
@lorenzo We should also update the [Docker-README](docker/README.md)

wevcode:
@lorenzo Update CI scripts. What do you think?

lorenzo:
Thanks, this looks great!

I will investagate at a later point if it is possible to consolidate the separate docker files into a single one, to speed the building process.

PR #221, opened 2018-06-16
Fix rule and dockerfile

lorenzo:
Consolidates the distributable docker containers into a single Dockerfile, that is actually buildable by docker hub.

Additionally fixed the multiple `CMD` and `ENTRYPOINT` rules, in the presence of multiple stages in the dockerfile


lorenzo:
I'll merge already so we can have a proper dockerfile

PR #223, opened 2018-06-26
Warn on plain text arguments for CMD and ENTRYPOINT

### LINKED ISSUE #198: Should suggest using the json array version for ENTRYPOINT and CMD
- [ ] This is a bug report

The json array version of `ENTRYPOINT` and `CMD` should be preferred over the shell version, as already recommended by the official guidelines: https://docs.docker.com/engine/reference/builder/#cmd

I've many hours debugging the different in behaviours between both versions.

lorenzo:
closes #198


lroellin:
Two nitpicks (I know I'm too late).

- Shouldn't this be called the _exec_ form? You can find this terminology [here](https://docs.docker.com/engine/reference/builder/#run)
- There's a typo (argume**ns** vs. argume**nts**)

lorenzo:
@lroellin it is actually referred in both terms, in the same page a few paragraphs below you can read:

> Note: In the JSON form, it is necessary to escape backslashes...

I personally prefer calling it the JSON form, as it actually gives you a hint of how it looks like.

Thanks for noticing the typo, I will fix it in the next release :)

PR #224, opened 2018-06-27
Added rule to warn on missing pipefail option

### LINKED ISSUE #170: Add a new rule for pipes in "RUN"
- [ ] This is a bug report

Add a new rule for pipes in `RUN`.

> https://github.com/docker/docker.github.io/blob/master/develop/develop-images/dockerfile_best-practices.md#using-pipes
#### Using pipes

Some `RUN` commands depend on the ability to pipe the output of one command into another, using the pipe character (`|`), as in the following example:

```Dockerfile
RUN wget -O - https://some.site | wc -l > /number
```

Docker executes these commands using the `/bin/sh -c` interpreter, which
only evaluates the exit code of the last operation in the pipe to determine
success. In the example above this build step succeeds and produces a new
image so long as the `wc -l` command succeeds, even if the `wget` command
fails.

If you want the command to fail due to an error at any stage in the pipe,
prepend `set -o pipefail &&` to ensure that an unexpected error prevents
the build from inadvertently succeeding. For example:

```Dockerfile
RUN set -o pipefail && wget -O - https://some.site | wc -l > /number
```

> **Note**: Not all shells support the `-o pipefail` option. In such
> cases (such as the `dash` shell, which is the default shell on
> Debian-based images), consider using the *exec* form of `RUN`
> to explicitly choose a shell that does support the `pipefail` option.
> For example:
>

```Dockerfile
RUN ["/bin/bash", "-c", "set -o pipefail && wget -O - https://some.site | wc -l > /number"]
```

lorenzo:
closes #170


zemanlx:
`/bin/sh: 1: set: Illegal option -o pipefail` I have hit this just two days ago

maybe we should recommend to include it `SHELL`, because than you do not need to do wraps in `RUN`... actually no change is needed

```dockerfile
SHELL ["/bin/bash", "-o", "pipefail", "-c"]
RUN apt-get update \
 && apt-get install --no-install-recommends -y \
    build-essential=12.3 \
```

it works aso with `alpine`'s `ash`
```bash

if ! (docker run --rm -it alpine /bin/ash -o pipefail -c "echo lol | grep no");then echo "Pipefail works";fi
Pipefail works
```

Also supported in `zsh`
Not supported in `dash` and `sh`

I am not using anything else. But I believe that `bash` and `ash` should cover most of usecases. Issue will be with `busybox:glibc` where you can find only `sh`.

lorenzo:
@zemanlx good idea, I'll make the rule so that it checks for `SHELL` with those options.

zemanlx:
We should also count on that it can look like 

```Dockerfile
SHELL ["/bin/bash", "-eo", "pipefail", "-c"]
SHELL ["/bin/bash", "-e", "-o", "pipefail", "-c"]
SHELL ["/bin/bash", "-o", "errexit", "-o", "pipefail", "-c"]
```
plus all possisble options and their permutations :wink:

lorenzo:
@zemanlx can you take a look again?

zemanlx:
@lorenzo Great. One last thing, we should fail on this check
```haskell
            it "warn on missing pipefail in the next image" $
                let dockerFile =
                        [ "FROM scratch as build"
                        , "SHELL [\"/bin/bash\", \"-o\", \"pipefail\", \"-c\"]"
                        , "RUN wget -O - https://some.site | wc -l file > /number"
                        , "FROM scratch as build2"
                        , "RUN wget -O - https://some.site | wc -l file > /number"
                        ]
                in ruleCatches usePipefail $ Text.unlines dockerFile
```

and if you can add following check for future me (it is passing)
```haskell
            it "warn on missing pipefail if next SHELL is not using it" $
                let dockerFile =
                        [ "FROM scratch as build"
                        , "SHELL [\"/bin/bash\", \"-o\", \"pipefail\", \"-c\"]"
                        , "RUN wget -O - https://some.site | wc -l file > /number"
                        , "SHELL [\"/bin/sh\", \"-c\"]"
                        , "RUN wget -O - https://some.site | wc -l file > /number"
                        ]
                in ruleCatches usePipefail $ Text.unlines dockerFile
```

src/Hadolint/Rules.hs
```diff
@@ -709,7 +710,29 @@ useJsonArgs = instructionRule code severity message check
...
+    check _ _ (Shell args) = (argumentsRule hasPipefailOption args, True)
+    check False _ (Run args) = (False, argumentsRule notHasPipes args)
+    check st _ _ = (st, True)
+    notHasPipes script = not (Bash.hasPipes script)
+    hasPipefailOption script =
+        not $
+        null
+            [ True
+            | cmd <- Bash.findCommands script
+            , validShell <- ["/bin/bash", "/bin/zsh", "bash", "zsh"]
```
zemanlx: Now I have just realized that Alpine `ash` should also be on the list.

lorenzo: alright, I will add it and then merge

lorenzo:
@zemanlx thanks for the great feedback, I fixed those 2 things you found

PR #228, opened 2018-07-04
Update version in docs & unify base image

zemanlx:
### What I did

- Updated version in docs
- Changed base image for Debian distro image to use same as we have for build. That should shorten a build time as we do not need to download another image.


lorenzo:
nice, thanks!

lorenzo:
feel free to merge whenever you like

PR #231, opened 2018-07-05
Warn on untrusted FROM registries

### LINKED ISSUE #225: [RFC] Warn if image from another registry is used in `FROM`
- [ ] This is a bug report

This feature is actually 2 features in one. The first feature is about automatically checking that a Dockerfile is does not use images from untrusted repositories.

Using the `FROM` instruction is a huge exercise in trust, you have to trust that a particular version or an image is safe for you to use, and that it will never be retagged maliciously. In order to prevent that, some companies copies trusted images into their own repositories, and reference them directly.

For example, this would be an untrusted image:

```dockerfile
FROM randomguy/fancy:10
...
```

But after an audit, the company decides to copy the image into their own repository, as it was deemed safe:

```dockerfile
FROM my-registry.com/fancy:10
...
```

The idea is that hadolint can warn whenever an untrusted repo is used. I propose the new `--trusted-repos` flag:

```bash
hadolint --trusted-repos my-registry.com Dockerfile
```

In order to support this feature, another feature needs to be implemented first: Optional rules. These type of rules won't run by default unless enabled by a flag or by a configuration file. The configuration file for this feature could for example be:

```yaml
ignored:
  - DL3002
trusted_repos:
  - my-registry.com
```

@zemanlx do you think hadolint is a good place to have this feature?

lorenzo:
### What I did

This implements and closes #225 

### How I did it

The CLI now accepts many `--trusted-registry` flags and a new config file key `trustedRegistries` to control what registries can be used in `FROM` instructions.

Example:

`hadolint --trusted-registry crazy.io --trusted-registry random.io Dockerfile`

### DIFF
--- a/src/Hadolint/Rules.hs
+++ b/src/Hadolint/Rules.hs
@@ -737,3 +737,15 @@ usePipefail = instructionRuleState code severity message check False
             , arg <- Bash.getAllArgs cmd
             , arg == "pipefail"
             ]
+
+allowedRegistry :: Set.Set Registry -> Rule
+allowedRegistry allowed = instructionRule code severity message check
+  where
+    code = "DL3026"
+    severity = ErrorC
+    message = "Only use one of the allowed registries in the FROM image"
+    check (From (UntaggedImage img _)) = Set.null allowed || isAllowed img
+    check (From (TaggedImage img _ _)) = Set.null allowed || isAllowed img
+    check _ = True
+    isAllowed Image {registryName = Just registry} = Set.member registry allowed
+    isAllowed Image {registryName = Nothing, imageName} = imageName == "scratch" || Set.member "docker.io"  allowed || Set.member "hub.docker.com" allowed
--- a/src/Hadolint/Rules.hs
+++ b/src/Hadolint/Rules.hs
@@ -10,7 +10,7 @@ import Data.Maybe (catMaybes)
 import qualified Hadolint.Bash as Bash
 import Language.Docker.Syntax
 
-import Data.Semigroup ((<>))
+import Data.Semigroup (Semigroup, (<>))
 import qualified Data.Set as Set
 import qualified Data.Text as Text
 import Data.Void (Void)
@@ -37,8 +37,20 @@ data RuleCheck = RuleCheck
     , success :: Bool
     } deriving (Eq)
 
+-- | Contains the required parameters for optional rules
+newtype RulesConfig = RulesConfig
+    { allowedRegistries :: Set.Set Registry -- ^ The docker registries that are allowed in FROM
+    } deriving (Show, Eq)
+
 instance Ord RuleCheck where
     a `compare` b = linenumber a `compare` linenumber b
+
+instance Semigroup RulesConfig where
+    RulesConfig a <> RulesConfig b = RulesConfig (a <> b)
+
+instance Monoid RulesConfig where
+    mempty = RulesConfig mempty
+    mappend = (<>)
 
 type IgnoreRuleParser = Megaparsec.Parsec Void Text.Text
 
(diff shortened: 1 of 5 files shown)


lorenzo:
I'll add docs to the readme in a bit

src/Hadolint/Rules.hs
```diff
@@ -737,3 +737,15 @@ usePipefail = instructionRuleState code severity message check False
             , arg <- Bash.getAllArgs cmd
             , arg == "pipefail"
             ]
+
+allowedRegistry :: Set.Set Registry -> Rule
+allowedRegistry allowed = instructionRule code severity message check
+  where
+    code = "DL3026"
+    severity = ErrorC
+    message = "Only use one of the allowed registries in the FROM image"
```
zemanlx: `Use only one ...` sounds better to me, if you do not mind.

Anyway, does it really allow only one registry from the list of registries? Why?

lorenzo: Well, you can only use one registry per `FROM` instruction. Maybe I should rename `one` with `an`:

`Only use an allowed registries in the FROM image`

zemanlx:
Test `Dockerfile`
```bash
$ cat Dockerfile.registry 
FROM ubuntu:18.04 AS builder1
FROM zemanlx/ubuntu:18.04 AS builder2
FROM docker.io/zemanlx/ubuntu:18.04 AS builder3
FROM good.io/ubuntu:18.04 AS builder4
FROM good.io/zemanlx/ubuntu:18.04 AS builder5
FROM bad.io/ubuntu:18.04 AS builder6
FROM bad.io/zemanlx/ubuntu:18.04 AS builder7
```

This accept images directly under `good.io` and also uder user `zemanlx` in registry `good.io`
```bash
$ hadolint --trusted-registry "good.io" Dockerfile.registry                               
Dockerfile.registry:1 DL3026 Only use one of the allowed registries in the FROM image
Dockerfile.registry:2 DL3026 Only use one of the allowed registries in the FROM image
Dockerfile.registry:3 DL3026 Only use one of the allowed registries in the FROM image
Dockerfile.registry:6 DL3026 Only use one of the allowed registries in the FROM image
Dockerfile.registry:7 DL3026 Only use one of the allowed registries in the FROM image
```

But that is not a case for `docker.io`
```bash
$ hadolint --trusted-registry "docker.io" Dockerfile.registry                             
Dockerfile.registry:2 DL3026 Only use one of the allowed registries in the FROM image
Dockerfile.registry:4 DL3026 Only use one of the allowed registries in the FROM image
Dockerfile.registry:5 DL3026 Only use one of the allowed registries in the FROM image
Dockerfile.registry:6 DL3026 Only use one of the allowed registries in the FROM image
Dockerfile.registry:7 DL3026 Only use one of the allowed registries in the FROM image
```

src/Hadolint/Rules.hs
zemanlx: You think is somebody will use `evil.io/docker.io/image`  with `--trusted-registry docker.io` and we will say it is ok because it is docker.io?

lorenzo:
@zemanlx I cannot reproduce that, I will add a test case to show you. This is the output from my shell:

```
cat foo
FROM ubuntu:18.04 AS builder1
FROM docker.io/zemanlx/ubuntu:18.04 AS builder3
```

```
hadolint --trusted-registry "random.io" foo
foo:1 DL3026 Only use one of the allowed registries in the FROM image
foo:2 DL3026 Only use one of the allowed registries in the FROM image
```

```
hadolint --trusted-registry "docker.io" foo
[no output]
```

src/Hadolint/Rules.hs
lorenzo: @zemanlx It's just bad wording, I pushed a change with something that I think reads better

zemanlx:
issue is in line 
```
FROM zemanlx/ubuntu:18.04 AS builder2
``` 
not in line 
```
FROM docker.io/zemanlx/ubuntu:18.04 AS builder3
```

lorenzo:
Ok, that’s a problem in the parser. I’ll add a workaround for the time being until I get around fixing that problem

lorenzo:
@zemanlx The bug is fixed now, after upgrading the parser

PR #233, opened 2018-07-06
More improvements to the shellcheck rule

### LINKED ISSUE #87: ShellCheck SC2154 on build ARG
```
RUN if [ ! "${http_proxy}" = "" ] ; then echo "Acquire::http::Proxy \"${http_proxy}\";" >> /etc/apt/apt.conf.d/98proxy; fi && \
    if [ ! "${https_proxy}" = "" ] ; then echo "Acquire::https::Proxy \"${https_proxy}\";" >> /etc/apt/apt.conf.d/98proxy; fi && \
    if [ ! "${http_proxy}" = "" ] ; then echo "http_proxy=${http_proxy}" >> /etc/wgetrc; fi && \
    if [ ! "${https_proxy}" = "" ] ; then echo "https_proxy=${https_proxy}" >> /etc/wgetrc; fi

```
Is raising a SC2154 http_proxy is referenced but not assigned. While this is appropriate if just looking at the shell script snippet. In the context of a Dockerfile this seems a bit too aggressive. Build time arguments will almost 100% be set outside.
Maybe it would be helpful to ignore such checks for variables declared as build arguments

### LINKED ISSUE #226: Change shell to real shell for ShellCheck

### What

We should also pass a value from `SHELL` to `shellcheck` in 
https://github.com/hadolint/hadolint/blob/7de84c2386b38234e9f6a01dd308b431d4404260/src/Hadolint/Bash.hs#L26 because current setting is hiding errors if shell is not `bash` which is mostly not true because [default is `sh`](https://docs.docker.com/engine/reference/builder/#shell).

### Steps to reproduce the behaviour

```Dockerfile
# Dockerfile.shell
FROM ubuntu:18.04 AS builder
SHELL ["/bin/bash", "-c"]
RUN echo $RANDOM
SHELL ["/bin/sh", "-c"]
RUN echo $RANDOM
```

```bash
$ docker build -t shelly -f Dockerfile.shell .
Sending build context to Docker daemon  2.953MB
Step 1/5 : FROM ubuntu:18.04 AS builder
 ---> 113a43faa138
Step 2/5 : SHELL ["/bin/bash", "-c"]
 ---> Running in 2689b9e0803c
Removing intermediate container 2689b9e0803c
 ---> f28dd8e83d5b
Step 3/5 : RUN echo $RANDOM
 ---> Running in 60458bd71409
13845
Removing intermediate container 60458bd71409
 ---> 33b682baec24
Step 4/5 : SHELL ["/bin/sh", "-c"]
 ---> Running in e1c238a26414
Removing intermediate container e1c238a26414
 ---> b0792e75154e
Step 5/5 : RUN echo $RANDOM
 ---> Running in 94dfc474759a

Removing intermediate container 94dfc474759a
 ---> a3e24dd32283
Successfully built a3e24dd32283
Successfully tagged shelly:latest

# Notice that $RANDOM is undefined in sh

$ hadolint Dockerfile.shell 
$
$ cat >shelly.sh                       
echo "$RANDOM"

# check script for /bin/sh
$ shellcheck -s sh shelly.sh           

In shelly.sh line 1:
echo "$RANDOM"
      ^-- SC2039: In POSIX sh, RANDOM is undefined.

# check script for /bin/bash
$ shellcheck -s bash shelly.sh
$
```

lorenzo:
### What I did
* Passing ARG and ENV as variables to the script, fixes #87 
* Using the detected SHELL as the start of the script, fixes #226

### DIFF
--- a/src/Hadolint/Bash.hs
+++ b/src/Hadolint/Bash.hs
@@ -5,7 +5,7 @@ module Hadolint.Bash where
 import Control.Monad.Writer (Writer, execWriter, tell)
 import Data.Functor.Identity (runIdentity)
 import Data.List (nub)
-import Data.Maybe (mapMaybe)
+import Data.Maybe (listToMaybe, mapMaybe)
 import Data.Semigroup ((<>))
 import qualified Data.Set as Set
 import qualified Data.Text as Text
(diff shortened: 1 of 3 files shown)


zemanlx:
Love it also with ARG and ENV. Do you think it would be possible to rename (`git mv`) `Bash.hs` to `Shell.hs` to remove confusion in future? Also in all other places where we mean `shell` but saying `bash`.

lorenzo:
@zemanlx I think that makes sense, will do that in a future cleanup PR, if that's ok with you

PR #254, opened 2018-08-12
Added a --config flag

### LINKED ISSUE #242: Config File path flag
Being able to point at a config flag (`--config` for example) in addition to the path search locations that are in place would be helpful.

Would simplify using hadolint in purpose-built containers

lorenzo:
This allows users to pass a custom confuguration file path.

closes #242

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,13 @@ file, in order or preference:
 
 In windows, the `%LOCALAPPDATA%` environment variable is used instead of `XDG_CONFIG_HOME`
 
+Additionally, you can pass a custom configuration file in the command line with
+the `--config` option
+
+```bash
+hadolint --config /file/to/config.yaml Dockerfile
+```
+
 ## Inline ignores
 
 It is also possible to ignore rules by using a special comment directly above the Dockerfile
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ Additionally, you can pass a custom configuration file in the command line with
 the `--config` option
 
 ```bash
-hadolint --config /file/to/config.yaml Dockerfile
+hadolint --config /path/to/config.yaml Dockerfile
 ```
 
 ## Inline ignores
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -46,6 +46,7 @@ data OutputFormat
 
 data LintOptions = LintOptions
     { showVersion :: Bool
+    , configFile :: Maybe FilePath
     , format :: OutputFormat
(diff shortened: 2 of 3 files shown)


app/Main.hs
```diff
@@ -80,13 +81,21 @@ parseOptions :: Parser LintOptions
...
     parseRulesConfig
   where
     version = switch (long "version" <> short 'v' <> help "Show version")
     --
+    -- | Parse the config filename to use
+    configFile =
+        optional
+            (strOption
+                 (long "config" <> short 'c' <> metavar "FILENAME" <>
+                  help "Use a custom file as configuration"))
```
zemanlx: Would you mind to change it to `Path to the configuration file`?

README.md
```diff
@@ -107,6 +107,13 @@ file, in order or preference:
 
 In windows, the `%LOCALAPPDATA%` environment variable is used instead of `XDG_CONFIG_HOME`
 
+Additionally, you can pass a custom configuration file in the command line with
+the `--config` option
+
+```bash
+hadolint --config /file/to/config.yaml Dockerfile
```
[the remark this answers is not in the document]
lorenzo: Intentional, you need to write up to the file name. The use case is that you may have different config files in the same folder.

app/Main.hs
lorenzo: It is not just the path to, it needs to include the file name

README.md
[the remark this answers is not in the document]
lorenzo: ah, I' m dumb... I understand what you mean now, will change

app/Main.hs
zemanlx: fair enough

lorenzo:
> I think that the is correct because we specify exactly one in CLI.

We can specify many docker registries in the cli, like we can specify many ignore rules

zemanlx:
@lorenzo yes, but only one per option
```bash
$ hadolint --trusted-registry lol.com phew.org docker/Dockerfile
hadolint: phew.org: openBinaryFile: does not exist (No such file or directory)

$ hadolint --trusted-registry lol.com --trusted-registry phew.org docker/Dockerfile
docker/Dockerfile:1 DL3026 Use only an allowed registry in the FROM image
docker/Dockerfile:31 DL3026 Use only an allowed registry in the FROM image
```

lorenzo:
@zemanlx I changed some help texts, is it better now?

PR #262, opened 2018-08-31
Add codacy formatter

jllopes:
Hey guys,

First of all, we really like the tool you’ve created, and we would love to support it on our product.
Because of this, we needed a formatter that would return an output matching the one we expect on our side, so we created one (Codacy.hs) similarly to the ones you already have on your tool and added it as an option to your Main.hs.

Let us know if something needs improvement.

Epic Code Reviews

### DIFF
--- a/src/Hadolint/Formatter/Codacy.hs
+++ b/src/Hadolint/Formatter/Codacy.hs
@@ -1,0 +1,73 @@
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE DeriveGeneric #-}
+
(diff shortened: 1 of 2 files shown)


src/Hadolint/Formatter/Codacy.hs
```diff
@@ -0,0 +1,71 @@
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE DeriveGeneric #-}
```
lorenzo: can you please remove this line?

lorenzo:
Thanks so much!

Issue #265, opened 2018-09-05
Access main functions and data structures and all rules programmatically


jllopes:
- [ ] This is a bug report
- [x] This is a feature request
- [x] I searched existing issues before opening this one

I'm trying to make a wrapper for Hadolint and I need access to some functions and data structures from the Main file (Lint and LintOptions). 
I've managed to do it but I had to do a refactor that renamed the current Main.hs so I could import the functions from that file to a different one.

So I was wondering if you guys would have any ideas on how to make the functions from the Main.hs file accessible in different files?

Also, is there a way to access all existing Hadolint rule names and descriptions programmatically to generate some JSONs?

lorenzo:
Just move the functions to a file inside the library (src folder) and call the functions from the Main file. I will merge the PR if you want to work on this.

lorenzo:
There is no way to access the descriptions programmatically at this point, other than doing some functional contorsions :)

PR #266, opened 2018-09-07
Refactor Main

jllopes:
Hey,

Following issue #265, I've refactored the code to make the lint function and LintOptions data structure available for files other than the main.
To do this I also had to move functions they depended on to another file.
Tell me if you think something needs to be redone.

Epic Code Reviews

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -5,36 +5,27 @@
 module Main where
 
 import Control.Applicative
-import Control.Monad (filterM)
-import Data.Coerce (coerce)
-import Data.Maybe (fromMaybe, listToMaybe)
+import qualified Data.List.NonEmpty as NonEmpty
 import Data.Semigroup ((<>))
 import qualified Data.Set as Set
 import Data.String
-import Data.Text (Text)
-import qualified Data.Yaml as Yaml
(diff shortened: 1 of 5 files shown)


lorenzo:
Do you mind if I push some comments directly to your branch?

jllopes:
@lorenzo Of course not, go ahead!

lorenzo:
@jllopes I did a bigger refactor, so that most of the Main file actually lives in the library

lorenzo:
@jllopes may I now ask what your use case was? Maybe the custom additions you want to create are also useful for everyone else and can be added directly to hadolint?

jllopes:
@lorenzo 
We want to support Hadolint on Codacy and for that we need a docker image that can read our config file and run the tool with what patterns we define on it.
I've just created a wrapper that parses that config file and then runs Hadolint programmatically based on the parsed configuration.
I don't know if it would be useful to add it directly to Hadolint, but the code will be on our public fork so you can check that out later and, if it makes sense to add it, I will create a PR

lorenzo:
cool, looking forward to see that. Thanks for your contributions!

PR #289, opened 2019-01-15
Update README with windows installation step

goostleek:
Thx to lukesampson/scoop#2972 it is possible to install `hadolint` using [scoop](https://github.com/lukesampson/scoop) installer.

### How to verify it


lorenzo:
Thanks!

PR #291, opened 2019-01-29
Update ShellCheck to 0.6.0

### LINKED ISSUE #287: pip version pin false positive
- [x] This is a bug report

### Expected behavior

A line like this:
    
    RUN pip install --index-url https://example.com/simple foobar==1.0.0

should be unproblematic.

### Actual behavior

    Dockerfile:3 DL3013 Pin versions in pip. Instead of `pip install <package>` use `pip install <package>==<version>`

See above.

```bash
Haskell Dockerfile Linter v1.15.0-0-g23c3bda
```

Obviously similar to #67, but that looks as though it should have been fixed long ago.

### LINKED ISSUE #290: hadolint fails when use apk add with --repository
- [x] This is a bug report

### Expected behavior
Run `apk add` with `--repository` argument.

### Actual behavior
Hadolint expects equal sign after `--repository` argument even if `apk` doesn't require it.
I have to place '=' after `--repository` to avoid `DL3018` error

```
FROM node:10.14.2-alpine
RUN apk add --no-cache --repository https://nl.alpinelinux.org/alpine/edge/testing flow=0.78.0-r0
```

```
DL3018 Pin versions in apk add. Instead of `apk add <package>` use `apk add <package>=<version>`
```

```bash
Haskell Dockerfile Linter v1.15.0-6-gaeb6041
```

zemanlx:
### What I did

- Updated SheckCheck from 0.5.0 to 0.6.0
- ￼Add allowed `repository` flag for `apk add` - fixes #290
- ￼Add allowed `index-url` falgs for `pip install` - fixes #287

### How I did it

- Updated stack to latest `lts` and fixed all issues that raised as `ShellCheck.Interface` was heavily refactored.

### How to verify it

Travis-CI should now pass for all environments, even `nightly`. For fixes, I have added tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,20 +75,24 @@ matrix:
       install:
         # Build image
         - travis_wait 30 docker build
-          --tag hadolint:$(git describe --tags --dirty)
-          --file docker/Dockerfile .
-        - docker build --tag hadolint:debian --file docker/Dockerfile --target debian-distro .
+          --tag hadolint:$(git describe --always --tags --dirty)
+          --file docker/Dockerfile
+          --target distro .
+        - docker build
+          --tag hadolint:$(git describe --always --tags --dirty)-debian
+          --file docker/Dockerfile
+          --target debian-distro .
       script:
         # List images
         - docker image ls
         # Lint its own Dockerfile
-        - docker run --rm -i hadolint:$(git describe --tags --dirty) <
+        - docker run --rm -i hadolint:$(git describe  --always --tags --dirty) <
           docker/Dockerfile
         # Check that version in hadolint in the same as its `git describe`
         - grep $(git describe --dirty) <<<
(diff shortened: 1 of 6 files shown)


src/Hadolint/Formatter/Format.hs
```diff
@@ -13,7 +13,6 @@ module Hadolint.Formatter.Format
 import Data.List (sort)
 import qualified Data.List.NonEmpty as NE
 import Data.Monoid (Monoid)
-import Data.Semigroup
```
lorenzo: This is required for compiling hadolint with olde GHC versions

lorenzo:
thanks for taking care of these outstanding things :)

src/Hadolint/Formatter/Format.hs
zemanlx: Ok I will put it back

PR #297, opened 2019-03-10
 Change the order of description of document's Docker rules in numerical order

zabio3:
### What I did

 -  Change the order of description of document's Docker rules in numerical order

#### Before
 - DL3005 -> DL3007 -> DL3006 -> DL3008

#### After
 - DL3005 -> DL3006 -> DL3007 -> DL3008


lorenzo:
Thanks, would you mind fixing the markdown listing error that was reported?

zabio3:
yes, please correct.

zabio3:
Thanks for the comment! I will fix the problem areas.

src/Hadolint/Rules.hs
```diff
@@ -447,8 +447,9 @@ aptGetPackages :: Shell.ParsedShell -> [String]
 aptGetPackages args =
     [ arg
     | cmd <- dropTarget <$> Shell.findCommands args
-    , arg <- Shell.getArgsNoFlags cmd
     , Shell.cmdHasArgs "apt-get" ["install"] cmd
+    , arg <- Shell.getArgsNoFlags cmd
+    , arg <- Shell.getArgsNoFlags cmd
```
lorenzo: This line is duplicated

lorenzo:
I'm sorry I did not notice the listing error was on the code and not on your change. There is no need for you to fix it now if you prefer. I can merge your original commit if you put your branch back to it

lorenzo:
Thanks

PR #298, opened 2019-03-12
use travis homebrew addon

stephengroat:
### What I did

update travis.yml to use addons

### How to verify it

travis build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,11 +53,12 @@ matrix:
     - env: *lts-fixed
       addons:
         artifacts: {paths: [./releases]}
+        homebrew:
+          update: true
+          packages: upx
       before_deploy:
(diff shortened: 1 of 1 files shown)


zemanlx:
@stephengroat thanks for the contribution!

Do you think that you can force it run only for deploy, because `brew` is time-consuming to run (it takes 386 seconds) and after your change, we will run it for every build.

zemanlx:
@stephengroat `v3.94` works fine, but if there is a way of using updated `v3.95` without `brew update` I would rather use that one.

stephengroat:
travis updates it `osx` images every once in a while, so you'll get that update eventually, but otherwise, there's no other easy/clean way that i know of (since `upx` doesn't release an `osx` binary)

lorenzo:
I think this makes sense as it is. thanks!

PR #302, opened 2019-03-14
Implement hlint suggestions

zemanlx:
### What I did
Moved guards forward as `hlint` was suggesting.

### How to verify it
`stack test`


lorenzo:
thanks!

PR #324, opened 2019-06-20
Add gem version pinning

stephengroat:
### What I did

add gem version pinning (to match pip, npm)

### How I did it

detect `gem install|i PACKAGE:VERSION`

### How to verify it

Tests included

### DIFF
--- a/src/Hadolint/Rules.hs
+++ b/src/Hadolint/Rules.hs
@@ -203,6 +203,7 @@ rules =
     , useJsonArgs
     , usePipefail
     , noApt
+    , gemVersionPinned
     ]
 
 optionalRules :: RulesConfig -> [Rule]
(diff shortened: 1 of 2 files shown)


lorenzo:
Thanks! This is super useful

SuperSandro2000:
We should probably add a small note that you wan't to move those gems into a Gemfile that is version pinned. Maybe the same for pips.

stephengroat:
@SuperSandro2000 `Gemfile` is associated with `bundler`, right? there would still need to be a `gem install bundler:VERSION` if bundler wasn't installed in the default ruby install.

Also, for installing a ruby script into `bin`, etc., I'm not sure that a `Gemfile` is better.

SuperSandro2000:
> Gemfile is associated with bundler, right?

yes

> there would still need to be a gem install bundler:VERSION if bundler wasn't installed in the default ruby install.

well, no. You would use ``apt install bundler`` or ``apk add ruby-bundler``

> Also, for installing a ruby script into bin, etc., I'm not sure that a Gemfile is better.

I can't come up with any scenario where you wouldn't want to use bundler, pip or npm to automatically install all your deps from a lock file instead of installing them manually.

stephengroat:
> well, no. You would use `apt install bundler` or `apk add ruby-bundler`

idk, not necessarily imo. with those package sources, there's the new dependency of `bundler` for everyone using `ruby` and the restriction of whatever version of the `bundler` gem is in the debian/alpine repo, etc. since a `Gemfile` is written with a specific version of `bundler`, some would probably want to pin specific versions of the `bundler` gem that may not available in those repos

> I can't come up with any scenario where you wouldn't want to use bundler, pip or npm to automatically install all your deps from a lock file instead of installing them manually.

this might be an example. if there's only a need to install 1 script-based gem (i.e. `terraforming`), making a `Gemfile` and installing & executing that with `bundle install && bundle exec terraforming` actually adds the dependency of `bundler`, which may not be a runtime dependency of the gem. also, that system may complicate the script since `bundle exec` creates it own subshell, etc.

SuperSandro2000:
Pip has even more of those tools...
But a suggestion to look into is it anyway as with more things to install it gets messy.

I always use rvm to install Ruby and it has bundler installed by default I think.

PR #327, opened 2019-06-22
Ignore stack lock file

weastur:
### What I did
Add to `.gitignore` stack lock file.


lorenzo:
Reading a bit more about this, I think we should commit the file instead. Did you have a reason for ignoring it?

weastur:
I thought you did. I'm not expert in Haskell and Stack, I've just seen artifacts after build.

lorenzo:
It's a new stack feature that I have not used, but I think it makes sense to commit the file. I'll do that shortly.

Thanks for your contribution anyways!

PR #333, opened 2019-07-03
Multiple fixes

### LINKED ISSUE #322: DL4001 when using multi-stage builds
- [x] This is a bug report

### Expected behavior
No issues should be raised

### Actual behavior
`[hadolint] Either use Wget or Curl but not both (DL4001)`

Create a Dockerfile with multi-stage build.
In the first image use: `curl` and `wget` in the second image.

#### Dockerfile: 
```
FROM debian:9.8 as go1_builder
LABEL product="gotest1"
ENV GO_VERSION="1.12.5"
WORKDIR /
RUN apt update && \
    apt install -y \
      wget \
      ca-certificates && \
      wget "https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz"

FROM debian:9.8 as go2_builder
LABEL product="gotest2"
ENV GO_VERSION="1.12.5"
WORKDIR /
RUN apt update && \ <--[hadolint] Either use Wget or Curl but not both (DL4001)
    apt install -y \
      ca-certificates \
      curl \
      libcurl4-gnutls-dev && \ <-- when && are removed, DL4001 is not reported anymore
      curl -sSLO "https://dl.google.com/go/go${GO_VERSION}.linux-amd64.tar.gz"
```

### LINKED ISSUE #325: False Positive for DL3028
- [x] This is a bug report

### Expected behavior
`RUN gem install bundler -v '2.0.2'` Throws no warnings

### Actual behavior
Specifying version with `-v` instead of `:` throws DL3028 warning.

1. Create a Dockerfile:
```Dockerfile
FROM ruby:2.6.3
RUN gem install bundler -v '2.0.2'
```
2. Run
```
hadolint Dockerfile
```
3. Result
```
Dockerfile:2 DL3028 Pin versions in gem install. Instead of `gem install <gem>` use `gem install <gem>:<version>`
```

```bash
Haskell Dockerfile Linter 1.17.0-no-git
```

**Dockerfile** (if relevant)

```Dockerfile
FROM ruby:2.6.3
RUN gem install bundler -v '2.0.2'
```

None

### LINKED ISSUE #332: False positive for DL3010 for non tar archives
Hello,

The current source code concerning DL3010 seems to consider that all files ending with:
https://github.com/hadolint/hadolint/blob/56fd3e1eed0e418b3a72dc619fc4282951d425e7/src/Hadolint/Rules.hs#L549
are tar archives.

The [Dockerfile ADD documentation](https://docs.docker.com/engine/reference/builder/#add) states:
> If `<src>` is a local tar archive in a recognized compression format...

I am not proficient in Haskell but a gzip/bzip2/xz archive is not necessarily a local tar archive.
Am I misunderstanding the behavior of this part of the source code?

Currently `hadolint` raises DL3010 when COPYing a bzip2 file even if this file is not a tar archive (and the Dockerfile ADD command will fail to extract such a file).

Referring to [the wikipedia tar page](https://en.wikipedia.org/wiki/Tar_(computing)#Suffixes_for_compressed_files), perhaps the right extensions to check would be something like:
```
archiveFormats = [".tar", ".tar.bz2", ".tb2", ".tbz", ".tbz2", ".tar.gz", ".tgz", ".tpz", ".tar.lz", ".tar.lzma", ".tlz", ".tar.xz", ".txz", ".tar.Z", ".tZ"]
```

It would be even better to check the real file format but I'm not sure if it is doable.

lorenzo:
fixes #332 
fixes #325 
fixes #322 

Additionally, change the YAML package for one implemented in pure Haskell. This will help me play with compiling hadolint to javascript

### DIFF
--- a/src/Hadolint/Rules.hs
+++ b/src/Hadolint/Rules.hs
@@ -546,7 +546,23 @@ useAdd = instructionRule code severity message check
             , format <- archiveFormats
             ]
     check _ = True
-    archiveFormats = [".tar", ".gz", ".bz2", "xz"]
+    archiveFormats =
+      [ ".tar"
+      , ".tar.bz2"
+      , ".tb2"
+      , ".tbz"
+      , ".tbz2"
+      , ".tar.gz"
+      , ".tgz"
+      , ".tpz"
+      , ".tar.lz"
+      , ".tar.lzma"
+      , ".tlz"
+      , ".tar.xz"
+      , ".txz"
+      , ".tar.Z"
+      , ".tZ"
+      ]
 
 invalidPort :: Rule
 invalidPort = instructionRule code severity message check
--- a/src/Hadolint/Rules.hs
+++ b/src/Hadolint/Rules.hs
@@ -326,6 +326,7 @@ wgetOrCurl = instructionRuleState code severity message check Set.empty
     severity = WarningC
     message = "Either use Wget or Curl but not both"
     check state _ (Run args) = argumentsRule (detectDoubleUsage state) args
+    check _ _ (From _) = withState Set.empty True -- Reset the state for each stage
     check state _ _ = withState state True
     detectDoubleUsage state args =
         let newArgs = extractCommands args
--- a/src/Hadolint/Rules.hs
+++ b/src/Hadolint/Rules.hs
@@ -847,7 +847,9 @@ gems args =
     [ arg
     | cmd <- Shell.findCommands args
     , Shell.cmdHasArgs "gem" ["install", "i"] cmd
+    , not (Shell.cmdHasArgs "gem" ["-v"] cmd)
     , arg <- Shell.getArgsNoFlags cmd
     , arg /= "install"
     , arg /= "i"
+    , not ("--" `isPrefixOf` arg)
     ]
--- a/test/Spec.hs
+++ b/test/Spec.hs
@@ -1066,6 +1066,35 @@ main =
                         ]
                 in ruleCatchesNot (registryIsAllowed ["random.com"]) $ Text.unlines dockerFile
         --
+        describe "Wget or Curl" $ do
+            it "warns when using both wget and curl" $
+                let dockerFile =
+                        [ "FROM node as foo"
+                        , "RUN wget my.xyz"
+                        , "RUN curl localhost"
+                        ]
+                in ruleCatches wgetOrCurl $ Text.unlines dockerFile
+            it "warns when using both wget and curl in same instruction" $
+                let dockerFile =
+                        [ "FROM node as foo"
+                        , "RUN wget my.xyz && curl localhost"
+                        ]
+                in ruleCatches wgetOrCurl $ Text.unlines dockerFile
+            it "does not warn when using only wget" $
+                let dockerFile =
+                        [ "FROM node as foo"
+                        , "RUN wget my.xyz"
+                        ]
+                in ruleCatchesNot wgetOrCurl $ Text.unlines dockerFile
+            it "does not warn when using both curl and wget in different stages" $
+                let dockerFile =
+                        [ "FROM node as foo"
+                        , "RUN wget my.xyz"
+                        , "FROM scratch"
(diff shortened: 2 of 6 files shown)


src/Hadolint/Rules.hs
```diff
@@ -546,7 +547,23 @@ useAdd = instructionRule code severity message check
...
+      , ".tar.gz"
+      , ".tgz"
+      , ".tpz"
+      , ".tar.lz"
+      , ".tar.lzma"
+      , ".tlz"
+      , ".tar.xz"
+      , ".txz"
+      , ".tar.Z"
+      , ".tZ"
```
SuperSandro2000: What about zip files?

lorenzo: do you know if `ADD` is able to extract zip files? I cannot remember

SuperSandro2000: No, it does not. 

Ref.:
````
If <src> is a local tar archive in a recognized compression format (identity, gzip, bzip2 or xz) then it is unpacked as a directory. Resources from remote URLs are not decompressed. When a directory is copied or unpacked, it has the same behavior as tar -x, the result is the union of:

Whatever existed at the destination path and
The contents of the source tree, with conflicts resolved in favor of “2.” on a file-by-file basis.
````
Source https://docs.docker.com/engine/reference/builder/#add

**Comment by lorenzo** [comment]
I'll merge now, since I'm just eager to use it at work

**Comment by DenysMorgun** [comment]
@lorenzo still getting 
```
DL3028 Pin versions in gem install. Instead of `gem install <gem>` use `gem install <gem>:<version>`
```
for 
```
RUN gem install nokogiri:1.7.2 -- \
        --use-system-libraries=true \
        --with-xml2-include=/usr/local/include/libxml2 \
--with-xml2-lib=/usr/local/lib
```
as mentioned here https://github.com/hadolint/hadolint/issues/325#issuecomment-505002856 :(

**Comment by lorenzo** [comment]
can you help identify how your case is different to this test case? https://github.com/hadolint/hadolint/pull/333/files#diff-3ae00181b173238649ff56154b9d01e4R183

File: test/Spec.hs [hunk 2]
```diff
@@ -173,6 +175,12 @@ main =
               it "pinned" $ do
                 ruleCatchesNot gemVersionPinned "RUN gem install bundler:1"
                 onBuildRuleCatchesNot gemVersionPinned "RUN gem install bundler:1"
+              it "does not warn on -v" $ do
+                ruleCatchesNot gemVersionPinned "RUN gem install bundler -v '2.0.1'"
+                onBuildRuleCatchesNot gemVersionPinned "RUN gem install bundler -v '2.0.1'"
```
biow0lf: `gem` also support `--version` as `-v`. Open new issue?

lorenzo: First confirm that you are using the latest version. I can't reproduce your case locally.

biow0lf:
```shell
$ hadolint --version
Haskell Dockerfile Linter 1.17.1-no-git
$ hadolint Dockerfile 
Dockerfile:5 DL3008 Pin versions in apt get install. Instead of `apt-get install <package>` use `apt-get install <package>=<version>`
Dockerfile:5 DL3009 Delete the apt-get lists after installing something
Dockerfile:15 DL3008 Pin versions in apt get install. Instead of `apt-get install <package>` use `apt-get install <package>=<version>`
Dockerfile:15 DL3009 Delete the apt-get lists after installing something
Dockerfile:34 DL3028 Pin versions in gem install. Instead of `gem install <gem>` use `gem install <gem>:<version>`
Dockerfile:66 DL3025 Use arguments JSON notation for CMD and ENTRYPOINT arguments
$
```

Link to Dockerfile: https://github.com/evemonk/evemonk/blob/master/Dockerfile

```shell
$ cat Dockerfile | grep "gem install"
RUN gem install bundler --version "$BUNDLER_VERSION" --force
$ 
```

I will open new issue.

PR #338, opened 2019-07-22
Add sh as valide shell for pipefail

william-billaud:
### What I did

Add the sh shell 9 and it absolute path /bin/sh to the list of valide shell for the rule DL4006

### How I did it

add "sh" and "/bin/sh"

### How to verify it

with this to dockerfile
```
FROM alpine:3.8

SHELL ["/bin/sh","-o", "pipefail","-c"]
RUN exit 1 | echo "bonjour" > file.txt
```
current output
```
hadolint.exe Dockerfile_nofail
Dockerfile_fail.txt:4 SC2216 Piping to 'echo', a command that doesn't read stdin. Did you want 'cat' instead?
Dockerfile_fail.txt:4 DL4006 Set the SHELL option -o pipefail before RUN with a pipe in it
```
and at build
```
docker build -t test -f Dockerfile_nofail .
Sending build context to Docker daemon  9.809MB
Step 1/3 : FROM alpine:3.8
 ---> dac705114996
Step 2/3 : SHELL ["/bin/sh","-o", "pipefail","-c"]
 ---> Using cache
 ---> 2cc990378d1d
Step 3/3 : RUN exit 1 | echo "bonjour" > file.txt
 ---> Running in 8c1239bbd4dd
The command '/bin/sh -o pipefail -c exit 1 | echo "bonjour" > file.txt' returned a non-zero code: 1
```

expected output :
```
hadolint.exe Dockerfile_nofail
Dockerfile_fail.txt:4 SC2216 Piping to 'echo', a command that doesn't read stdin. Did you want 'cat' instead?
```

### DIFF
--- a/src/Hadolint/Rules.hs
+++ b/src/Hadolint/Rules.hs
@@ -801,7 +801,7 @@ usePipefail = instructionRuleState code severity message check False
         null
             [ True
             | cmd@(Shell.Command name arguments _) <- Shell.presentCommands script
-            , validShell <- ["/bin/bash", "/bin/zsh", "/bin/ash", "bash", "zsh", "ash"]
+            , validShell <- ["/bin/bash", "/bin/zsh", "/bin/ash","/bin/sh", "bash", "zsh", "ash","sh"]
             , name == validShell
             , Shell.hasFlag "o" cmd
             , arg <- Shell.arg <$> arguments


lorenzo:
Tests are failing, can you take a look at what needs to be fixed?

SuperSandro2000:
The tests check for this to fail.
See https://travis-ci.org/hadolint/hadolint/jobs/562155365#L566

zemanlx:
@william-billaud in `alpine` please use `ash`.
Linux `sh` is not valid shell

```bash
$ docker run --rm -it ubuntu sh -o pipefail
sh: 0: Illegal option -o pipefail
```

SuperSandro2000:
```shell
$ docker run --rm -it alpine sh -o pipefail
/ #
```
Busybox sh does support pipefail or does not error at all.

zemanlx:
In `alpine` almost everything basic is `busybox` which implements `-o pipefail` so it can be confusing. That's why we added `ash` (alpine specific) to valid shells list and not general `sh`. 

```
$ docker run --rm -it alpine:3.8 ash       
/ # ls -lah /bin/sh
lrwxrwxrwx    1 root     root          12 Mar  6 20:15 /bin/sh -> /bin/busybox
/ # ls -lah /bin/ash
lrwxrwxrwx    1 root     root          12 Mar  6 20:15 /bin/ash -> /bin/busybox
```

SuperSandro2000:
Then sh should not be valid for ``-o pipefail`` except if you are on alpine.
So in theory it should suggest to use ash and not sh cause otherwise it gets really confusing.

zemanlx:
@william-billaud thanks, for your work! Unfortunately, I am closing this PR as it can mislead users to errors we should catch.

zemanlx:
@SuperSandro2000 we are mentioning the usage of `ash` for `alpine` in wiki for [DL4006](https://github.com/hadolint/hadolint/wiki/DL4006). Feel free to enhance it so it is clear.

SuperSandro2000:
> @SuperSandro2000 we are mentioning the usage of `ash` for `alpine` in wiki for [DL4006](https://github.com/hadolint/hadolint/wiki/DL4006). Feel free to enhance it so it is clear.

I think the wiki is good enough if you read it thoughtful.

PR #351, opened 2019-10-01
Support pip requirement pinning with long parameter

x-way:
### What I did
When using the long parameter for installing pip packages (`RUN pip install --requirement /tmp/requirements.txt`), this get's incorrectly reported by hadolint as un-pinned package versions, as hadolint currently only recognizes the short parameter (`RUN pip install -r /tmp/requirements.txt`).
This PR extends the parsing rules so that also the long parameter is recognized properly.

### How to verify it

When having a `RUN pip install --requirement /tmp/requirements.txt
` statement in the Dockerfile, previously hadolint reported:
```
Dockerfile:4 DL3013 Pin versions in pip. Instead of `pip install <package>` use `pip install <package>==<version>`
```
With the changed version, it correctly recognizes it as an installation of pinned package versions.

### DIFF
--- a/src/Hadolint/Rules.hs
+++ b/src/Hadolint/Rules.hs
@@ -595,7 +595,7 @@ pipVersionPinned = instructionRule code severity message check
     -- in running this rule
     relevantInstall cmd =
         ["install"] `isInfixOf` Shell.getArgs cmd &&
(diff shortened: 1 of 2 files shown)


lorenzo:
Thanks!

PR #352, opened 2019-10-03
Update shellcheck

zemanlx:
### What I did

Updated stack to latest `lts`, updated `shellcheck` to `0.7.0` and fixed `hlint` complains.

### How I did it

The best I can.

### How to verify it

CI should pass.


lorenzo:
thanks!

Issue #250, opened 2018-08-01
 DL4006 with BusyBox ash


vitaliytv:
- [x] This is a bug report
- [ ] This is a feature request
- [x] I searched existing issues before opening this one

### Expected behavior

When using SHELL ["/bin/sh", "-o", "pipefail", "-c"], the commands in a RUN should not be fail with DL4006

### Actual behavior

/dev/stdin:26 DL4006 Set the SHELL option -o pipefail before RUN with a pipe in it
Exited with code 1

### Steps to reproduce the behavior

Last docker version

```bash
Haskell Dockerfile Linter v1.10.4-3-g20c5c3b
```

Dockerfile for reproduce:

```Dockerfile
FROM alpine:3.8

SHELL ["/bin/sh", "-o", "pipefail", "-c"]

RUN curl https://get.acme.sh | sh
```

zemanlx:
That's tricky. `hadolint` have no idea that `/bin/sh` is pointing to `/bin/busybox` which supports `-o pipefail`. 

I am not able to image how we can solve this. I would recommend to ignore it in comment. Any other ideas @lorenzo?

matrixik:
Why not using:
```
SHELL ["/bin/ash", "-eo", "pipefail", "-c"]
```

vitaliytv:
@matrixik  offer solution, thanks

PR #365, opened 2019-11-18
Adds `/bin/sh` as a valid shell for DL4006

sdwolfz:
### What I did
- Added `/bin/sh` and `sh` as valid shells that can be linted for `-o pipefail` in DL4006
- Fixed the regression test.
- Added a stack generated file to the gitignore.

### How I did it
I noticed that a Dockerfile with
```dockerfile
SHELL ["/bin/sh", "-o", "pipefail"]
RUN env | grep FOO
```
Was failing `DL4006`.

So I checked the linter code and `/bin/sh` and `sh` were not present in the list of valid shells. I added them there and fixed the tests.

### How to verify it

Run `hadolint` on the snippet above before and after the change. Or check the tests.


lorenzo:
Thanks a lot for the contribution, but regular `sh` does not actually allow `pipefail` :(

If you are using `alpine`, please follow this suggestion instead https://github.com/hadolint/hadolint/issues/250#issuecomment-409891140

sdwolfz:
Hmm, in that case are you ok with updating the message text to hint at either chaining the shell in case it's sh or disabling the check?

lorenzo:
@sdwolfz sure, I'm ok with updating the text. Would you like to do that?

sdwolfz:
Another idea would be to detect 'set -o pipefail' in the `RUN` itself before a pipe is used, but I don't have the knowledge to implement that right now.

sdwolfz:
> @sdwolfz sure, I'm ok with updating the text. Would you like to do that?

Yes I'll get to it in a few hours.

sdwolfz:
@lorenzo The message is updated, please take a look :pray:

lorenzo:
Thanks!!

PR #377, opened 2020-01-01
alpine image

aogier:
### What I did

I've added an alpine target image, updating travis and documentation stuff where necessary. This come after some weeks of tests in my company where we need to use hadolint on jenkins' k8s plugin and we use alpine for the sake of image size (~8MiB vs debian's 53). We'd be glad if you consider adopting this enhancement.

### How I did it

Updated READMEs, travis config, dockerfile and his hooks. I didn't touched INTEGRATION docs because I'm unsure on how to proceed w/o rendering that file too much "noisy", what do you think about?

### How to verify it

Pipelines should pass

HTH, ciao!

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -35,3 +35,7 @@ CMD ["/bin/hadolint", "-"]
 FROM scratch AS distro
 COPY --from=builder /root/.local/bin/hadolint /bin/
 CMD ["/bin/hadolint", "-"]
+
+FROM alpine:3 AS alpine-distro
+COPY --from=builder /root/.local/bin/hadolint /bin/
+CMD ["/bin/hadolint", "-"]
--- a/.travis.yml
+++ b/.travis.yml
@@ -81,6 +81,10 @@ matrix:
           --tag hadolint:$(git describe --always --tags --dirty)-debian
           --file docker/Dockerfile
           --target debian-distro .
+        - docker build
+          --tag hadolint:$(git describe --always --tags --dirty)-alpine
+          --file docker/Dockerfile
+          --target alpine-distro .
       script:
         # List images
         - docker image ls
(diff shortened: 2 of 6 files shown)


zemanlx:
Thank you @aogier! Really nice PR. I remember that we were trying to get `alpine` build, but there wasn't support for that in `stack`. Copying `glibc` binary to `alpine` is an interesting workaround. Still not sure if there aren't any `glibc`/`musl-libc` issues. Do you use it in production/CI?

aogier:
Hi @zemanlx, I can confirm I'm using an alpine version in production since a while w/o any problems, my current container is substantially the same submitted in this PR:

```dockerfile

ARG HADOLINT_VERSION=latest
ARG ALPINE_VERSION=latest
FROM hadolint/hadolint:${HADOLINT_VERSION} as source

FROM alpine:${ALPINE_VERSION}

COPY --from=source /bin/hadolint /bin
```

Rationale behind my repackaging choice was that we have a statically linked binary that doesn't use too many syscalls so libc/musl issues shouldn't happen here... OTOH I don't know haskell so apart this consideration my experience is pretty empirical but so far so good ! :)

Let me know if I can help you in any way, and thank you, ciao

zemanlx:
OK, we can give it a shot. Thanks!

docker/Dockerfile
```diff
@@ -35,3 +35,7 @@ CMD ["/bin/hadolint", "-"]
 FROM scratch AS distro
 COPY --from=builder /root/.local/bin/hadolint /bin/
 CMD ["/bin/hadolint", "-"]
+
+FROM alpine:3 AS alpine-distro
+COPY --from=builder /root/.local/bin/hadolint /bin/
```
SuperSandro2000: According to HFS it would belong to `/usr/bin/`

PR #378, opened 2020-01-01
added jenkins' k8s plugin example

aogier:
### What I did

Added jenkin's k8s plugin example in integrations docs


zemanlx:
Thank you!

PR #392, opened 2020-02-11
On format Checkstyle exit with success on findings

### LINKED ISSUE #391: Suppress Exit Code
- [ ] This is a bug report

### Expected behavior

When running in a CICD environment, I would like to suppress the exit code on findings. That way I would know when the command fails due to wrong usage (file missing) and could evaluate the results that are written to a checkstyle result file later on in Jenkins / SonarQube. 
I would like to pass something like `--suppress-exit-code`

### Actual behavior

Exit code is `1` if I have finding and command execution issues.

CCFenner:
fixes #391

### What I did

Just copy&paste from `codacy` implementation.

### How to verify it

Use a `Dockerfile` with issues like this:

```Dockerfile
FROM gcr.io/distroless/base:latest
```

```sh
hadolint ./Dockerfile --format checkstyle > /dev/null && echo OK || echo FAILED
```

### DIFF
--- a/src/Hadolint/Lint.hs
+++ b/src/Hadolint/Lint.hs
@@ -45,7 +45,7 @@ printResultsAndExit format allResults = do
         case format of
             TTY -> TTY.printResult res
             Json -> Json.printResult res
-            Checkstyle -> Checkstyle.printResult res
+            Checkstyle -> Checkstyle.printResult res >> exitSuccess
             CodeclimateJson -> Codeclimate.printResult res >> exitSuccess
             Codacy -> Codacy.printResult res >> exitSuccess
 


lorenzo:
I was reading about `checkstyle` and apparently CI systems using it expect the exit code to be the number of warnings found. While we are not following it (we always return as if we only found 1 error), I think the current behavior is more correct.

What do you think?

CCFenner:
In this case only the output format is used, not checkstyle itself. So I think bringing in consistency by aligning the exit codes when file output is used, would be fine.

The issue I have is that I would like to distinguish between processing issues and static check issues, see #391. 
Would it be possible to add a flag disabling exit codes resulting from findings?

lorenzo:
@CCFenner It would be possible to add a flag, certainly. I wonder if using a bash solution might be sufficient in your case:

```sh
hadolint ./Dockerfile --format checkstyle || true
```

that will make the command succeed every time and will still output the results to stdout.

zemanlx:
Unfortunately, I also cannot support this idea. It is changing current behaviour and therefore it can break other people pipelines which not acceptable.

However, I like the idea of a flag that will suppress non-zero code for checks that you have suggested in #391.

PR #416, opened 2020-04-20
Use lts-15.3

markus1189:
### What I did
Update to stackage lts-15.3

### How I did it
Update in stack.yaml, then fix everything that was broken.

### How to verify it
Package builds


zemanlx:
That's great @markus1189! I have tried to upgrade `stack` but my limited knowledge of Haskell was holding me back. Thank you for your contribution!

PR #417, opened 2020-04-21
Add cabal flag for static compilation

markus1189:
Adds a flag to the hadolint package that toggles static compilation.
By default I set it to `false` and enable this flag in the `Dockerfile`, 
`appveyor.yml` and `travis.yml`.

The reason is that some systems don't easily support static linking,
so using the flag when you want it is nicer instead of having to patch
around this problem.

I tested locally that it "works" without setting anything and that it "breaks" (for me on NixOS) when enabling the `static` flag.

We could also set the flag to `true` by default, and I would manually patch the package for NixOS, but it feels like disabling the static compilation by default is a saner default.  WDYT?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -141,7 +141,7 @@ before_install:
 
 install:
   - travis_retry stack --no-terminal --install-ghc $ARGS test --only-dependencies
-  - stack --no-terminal $ARGS install --ghc-options='-fPIC'
+  - stack --no-terminal $ARGS install --ghc-options='-fPIC' --flag hadolint:static
 
 script:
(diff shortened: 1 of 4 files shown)


zemanlx:
Thank you for your contribution! The idea behind using static binary for hadolint is that we do not have a list of dependencies/issues which are required for people to use hadolint. They can simply download binary and use it.

If I understand it correctly, your change is not building static binaries if you run it locally using stack but it still does this for our releases and CI. That sounds OK to me. What do you think @lorenzo?

lorenzo:
look good. Thanks!

PR #423, opened 2020-05-07
Add hadolint-shim

kornicameister:
This is somehow new but I've been relying on such solution for longer time and it proven to be quite robust used both in shell and `neovim` integration.

### What I did

Created an alternative approach to use `hadolint` via shim that gets installed somewhere into `$PATH`

### How I did it

Wrote simple [bin/hadolint](https://github.com/kornicameister/dotfiles/commits/master/bin/hadolint) shim longer while ago and have decided to extract it [here](https://github.com/kornicameister/hadolint-shim) for others to pick up if needed. There are editor integrations for sure but sometimes there is need to just do `hadolint` in `$SHELL` and it is much simpler to just do `hadolint Dockerfile` instead of `docker run ...` on systems that do not provide static binaries.

### How to verify it

Following installation instruction [here](https://github.com/kornicameister/hadolint-shim/blob/master/README.md#installation). 

PS. I will provide basic CI later on will comment on this PR as well. This should provide a good verification example.

### DIFF
--- a/docs/INTEGRATION.md
+++ b/docs/INTEGRATION.md
@@ -261,9 +261,13 @@ if docker run --rm -i hadolint/hadolint < "%d/%f"
 | grep -EC100 ':WARNING:' ; then exit 1 ; else exit 0 ; fi
 ```
 
-## Shims
+## Other
 
-- [hadolint-shim](https://github.com/kornicameister/hadolint-shim)
+Following section contains different angles on `hadolint` usage that go outside of [code review](#code-review) or 
+[editors' integration](#editors). 
+
+- [hadolint-shim](https://github.com/kornicameister/hadolint-shim) - use `hadolint` as normal executable on systems where
+  static binaries are not accessible
 
 [linter-hadolint]: https://atom.io/packages/linter-hadolint
 [linter-hadolint-img]: https://user-images.githubusercontent.com/18702153/33764234-7abc1f24-dc0b-11e7-96b6-4f08207b6950.png
--- a/docs/INTEGRATION.md
+++ b/docs/INTEGRATION.md
@@ -267,7 +267,7 @@ Following section contains different angles on `hadolint` usage
 that go outside of [code review](#code-review) or [editors' integration](#editors). 
 
 - [hadolint-shim](https://github.com/kornicameister/hadolint-shim) - use `hadolint` 
-  as normal executable on systems where static binaries are not accessible
+  as normal executable on systems where static binaries are not available
 
 [linter-hadolint]: https://atom.io/packages/linter-hadolint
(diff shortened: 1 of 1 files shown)


docs/INTEGRATION.md
```diff
@@ -261,6 +261,10 @@ if docker run --rm -i hadolint/hadolint < "%d/%f"
 | grep -EC100 ':WARNING:' ; then exit 1 ; else exit 0 ; fi
 ```
 
+## Shims
+
+- [hadolint-shim](https://github.com/kornicameister/hadolint-shim)
```
lorenzo, line 266: Could you add some text after the header explaining what it does? I might be helpful to others

kornicameister: @lorenzo I've changed `## Shims` to `## Others`. If there will be more of shims later on,  documentation can be changed, I believe. For now my little `shim` lands in **other** just because there is nothing like that elsewhere (at least I am not aware of it) + using term `shim`, given its definition, might be a little exaggerating ;-)

docs/INTEGRATION.md
```diff
@@ -261,6 +261,14 @@ if docker run --rm -i hadolint/hadolint < "%d/%f"
 | grep -EC100 ':WARNING:' ; then exit 1 ; else exit 0 ; fi
 ```
 
+## Other
+
+Following section contains different angles on `hadolint` usage 
+that go outside of [code review](#code-review) or [editors' integration](#editors). 
+
+- [hadolint-shim](https://github.com/kornicameister/hadolint-shim) - use `hadolint` 
+  as normal executable on systems where static binaries are not accessible
```
SuperSandro2000, line 270: Maybe it should be added that it uses docker and why you want to use that but I am not to sure.

kornicameister: It might be, not saying that it shouldn't. But if we do that here, shouldn't similar explanation go to each entry.  For instance, you have `ale` for `vim` that I am using as well but there's no such explanation there + and it _can_ use `docker` variant too. I mean, mine is using `docker`, other integration might use that, but how `hadolint` is used in specific integration goes a little bit outside of scope of `hadolint` itself, doens't? What is important is that they make life easier.

kornicameister: Though, maybe `accessible` ought to be changed to `available` to emphasize that binaries are available only for certain platforms as mentioned [here](https://github.com/hadolint/hadolint#install).

lorenzo: I agree @kornicameister

docs/INTEGRATION.md
lorenzo: sounds good to me :)

kornicameister:
@SuperSandro2000 @lorenzo just made that little rewording. Hope everything's in order now :+1:

lorenzo:
thanks!

PR #440, opened 2020-06-16
Added new rule DL3029 to check platform flag in FROM

### LINKED ISSUE #429: Add a new rule to detect supported architectures of a dockerfile
Hi Team,

I am interested in adding a support in hadolint tool for detecting if the dockerfile is architecture-specific.
ex : If my base image in the FROM clause is aarch64 architecture-specific, and I am linting that dockerfile on x86_64 platform, then a rule should pop up saying that “the base image is not supported on this architecture”.

We can use the "docker manifest inspect" experimental command to fetch architecture information from the image's manifest.

I have tested the same in the "main" module defined in Main.hs file, and am able to get the required result there.

I suggest that maybe we can add a new rule, by adding a similar configuration in Rules.hs to detect the supported architectures of a dockerfile.

Please share your opinion on the same.

odidev:
@lorenzo ,
Added new rule **DL3029** to check **--platform** flag in `FROM`. 

Modified README.md and Rules.hs.

Closes #429.

Please review and provide feedback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,6 +191,7 @@ Please [create an issue][] if you have an idea for a good rule.
 | [DL3026](https://github.com/hadolint/hadolint/wiki/DL3026)   | Use only an allowed registry in the FROM image                                                                                                      |
 | [DL3027](https://github.com/hadolint/hadolint/wiki/DL3027)   | Do not use `apt` as it is meant to be a end-user tool, use `apt-get` or `apt-cache` instead                                                         |
 | [DL3028](https://github.com/hadolint/hadolint/wiki/DL3028)   | Pin versions in gem install. Instead of `gem install <gem>` use `gem install <gem>:<version>`                                                       |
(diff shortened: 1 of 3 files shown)


zemanlx:
@odidev Thank you for your contribution! Would you mind also add a few tests into https://github.com/hadolint/hadolint/blob/master/test/Spec.hs?

odidev:
> @odidev Thank you for your contribution! Would you mind also add a few tests into https://github.com/hadolint/hadolint/blob/master/test/Spec.hs?

Thanks for the feedback.

I have added two tests for testing the rule. Please review and let me know if any more changes required.

zemanlx:
Great! Thank you also for creating a Wiki page for it.

odidev:
Thanks @zemanlx for approval and merging the rule.

lorenzo:
Thanks @odidev

PR #456, opened 2020-08-14
Support --trusted-host for pip install

davidhart82:
Fixes: #459

Fixes false-positive for `pipVersionNotPinned` when using `--trusted-host`; new test added.


davidhart82:
Build seems to be failing globally.

davidhart82:
This fixes #459

lorenzo:
thanks!

PR #465, opened 2020-10-03
Makefile: Initial commit

Kreyren:
This merge requests adds a makefile allowing for more efficient management of the repository as build instruction can be standardized and the make used for optimizing CI.

Signed-off-by: Jacob Hrbek <kreyren@rixotstudio.cz>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@ stack.yaml.lock
 
 # travis wait is generating log in PWD
 travis_wait_*.log
+
+# Ignore build target for git
+build/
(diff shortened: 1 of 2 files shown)


zemanlx:
Thank you @Kreyren for the contribution. Can you please elaborate on why using plain `stack` is not enough? I would like to avoid adding another dependency to build process if it does not come with great benefits.

lorenzo:
Thanks @Kreyren, we would rather not have this for the time being as we already are using a build tool.

Issue #470, opened 2020-10-05
Lint rpm based package managers: dnf, yum, zypper


m-ildefons:
- [ ] This is a bug report
- [x] This is a feature request
- [x] I searched existing issues before opening this one

Hadolint is a great tool, but unfortunately only supports linting for Debian- and Alpine-like package managers.
It would be a great improvement to be able to lint `yum`, `dnf` and `zypper` commands like `apt` commands:
  * No update/upgrade/dist-upgrade.
  * Use `-y` for yum, dnf and `-n` for zypper or these commands may fail.
  * Clean caches, e.g. `yum clean all` should end an installation command.

### Expected behavior

Linter complains about wrong application of `yum`:
`/dev/stdin:2 DLXXXX Do not use yum update`
`/dev/stdin:2 DLXXXX Do not use yum without '-y' flag.`

### Actual behavior

No output. Linter pretends Dockerfile is fine.

### Steps to reproduce the behavior

**Output of `hadolint --version` or
  `docker run --rm hadolint/hadolint hadolint --version`:**

```bash
Haskell Dockerfile Linter v1.18.0-15-g6664056
```

**Dockerfile** (if relevant)

```Dockerfile
FROM centos:8
RUN yum update
```

**Additional environment details (OS, stack version, etc.)**
Linux 4.19.0-10-amd64 #1 SMP Debian 4.19.132-1 (2020-07-24) x86_64 GNU/Linux
Docker version 18.09.1, build 4c52b90

code-noggin:
Why is `yum update` bad? I often have a base image with some core packages that need to be updated with security fixes.

m-ildefons:
@code-noggin the same reasoning as with apt-get applies:
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get
Updating core packages can fail, especially in unprivileged containers.

In addition to the reason from the docker docs, `yum update` breaks reproducibility of images.
There is the scenario, where you build an app now and it works fine even with the `yum update` command in the Dockerfile, but 6 months from now, you make a change and therefore rebuild the Docker image and suddenly the image no longer works because the `yum update` pulled a newer, incompatible version of some library or something.

Maybe consider using `yum update-to -y foobar-x.y.z` or `yum install -y foobar-x.y.z` to fix the security issues?
This should avoid all problems I can think of right now, as you would need to name the packages and versions you want to update to explicitly.

zemanlx:
[...]

PR #471, opened 2020-10-06
Rules: Add rules for `yum` commands, #470

m-ildefons:
Add four rules for `yum` commands warning on bad practices:

  * `yum update`
  * `yum install` without non-interactive flag `-y`
  * `yum install -y <pkg>` without version pinning
  * `yum install ...` without also cleaning up with `yum clean all`

These four rules are accompanied by test for each rule.
Also included is a new helper function `Shell.anyCommands`.

### What I did

Add four rules for linting `yum` commands in `RUN` layers. These four rules check for common mistakes and bad-practices with `yum` in Dockerfiles.
This PR is related to #470, but does not suffice to close it since `zypper` and `dnf` checks are still missing.

### How I did it

I expanded the `Rules.hs` file. I also added a new helper function in the `Shell.hs` file: `Shell.anyCommands` that works like `Shell.allCommands`, except it matches on any instead of all commands in a shell line.

### How to verify it

Run the test suite, or run `hadolint` on this Dockerfile:

```Dockerfile
FROM centos:8
RUN yum update
RUN yum install httpd
RUN yum install -y tomcat
RUN yum install -y mariadb-10.4
RUN yum install -y php-7.2 && yum clean all
```
The `FROM` and the last `RUN yum install..` line will not produce warnings and the other lines will each trigger a different combination of warnings for `yum`.

### DIFF
--- a/src/Hadolint/Rules.hs
+++ b/src/Hadolint/Rules.hs
@@ -205,6 +205,10 @@ rules =
     , usePipefail
     , noApt
     , gemVersionPinned
+    , yumYes
+    , noYumUpdate
+    , yumCleanup
+    , yumVersionPinned
     ]
 
 optionalRules :: RulesConfig -> [Rule]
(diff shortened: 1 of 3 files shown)


lorenzo:
looking great! I wonder why the travis build did not start

lorenzo:
some tests are failing too, but they seem easy to fix

m-ildefons:
Thanks. I fixed the overzealous `clean all` rule. Now it should no longer falsely flag `RUN` layers that don't even contain `yum` commands.

lorenzo:
@m-ildefons Would you mind adding the corresponding wiki pages for the new `DL` codes? I think they can just be compies of the similar codes for `apt`

lorenzo:
Thanks so much for your contribution!

src/Hadolint/Rules.hs
```diff
@@ -877,6 +881,62 @@ noPlatformFlag = instructionRule code severity message check
...
+    check _ = True
+    yumInstall = Shell.cmdHasArgs "yum" ["install"]
+    yumClean = Shell.cmdHasArgs "yum" ["clean", "all"]
+
+yumVersionPinned :: Rule
+yumVersionPinned = instructionRule code severity message check
+  where
+    code = "DL3033"
+    severity = WarningC
+    message = "Sepcify version with `yum install -y <package>-<version>`."
```
lorenzo, line 928: ah, I merged too quickly and I just found this is a typo `Sepcify` @m-ildefons

Issue #452, opened 2020-07-29
Basic auth in index-url for pip fails rule


bneijt:
- [x] This is a bug report
- [x] I searched existing issues before opening this one

### Expected behavior

   pip install --index-url https://user:pass@example.com arrow==10

should be valid as I pin the version.

### Actual behavior

Rule `DL3013` is triggered.

### Steps to reproduce the behavior

Create a dockerfile with

  RUN pip install --index-url https://user:pass@example.com arrow==1.0.0

Execute hadolint on the docker file.

**Output of `hadolint --version` or
  `docker run --rm hadolint/hadolint hadolint --version`:**

```bash
% stack run -- hadolint --version                                                                                                                                                                         
Haskell Dockerfile Linter v1.18.0-6-ga0d655d
```

**Additional environment details (OS, stack version, etc.)**
Linux,
stack Version 2.3.1, Git revision de2a7b694f07de7e6cf17f8c92338c16286b2878 (dirty) (8103 commits) x86_64

Also reproducable in spec test:
```haskell
            it "pip install --index-url with --extra-index-url with basic auth" $ do
                ruleCatchesNot
                    pipVersionPinned
                    "RUN pip install --index-url https://user:pass@eg.com/foo --extra-index-url https://user:pass@ex-eg.io/foo foobar==1.0.0"
                onBuildRuleCatchesNot
                    pipVersionPinned
                    "RUN pip install --index-url https://user:pass@eg.com/foo --extra-index-url https://user:pass@ex-eg.io/foo foobar==1.0.0"
```

davidhart82:
Can confirm that we see this too.

PR #473, opened 2020-10-06
Python pip basic auth | fixes [#452]

dariodsa:
### What I did
This PR resolves issue #452. Problem was in wrong arguments ordering which was created by ShellCheck. For example, pip command mentioned in issue was cut down to:
```
[
CmdPart {arg = "install", partId = 6},
CmdPart {arg = "--index-url", partId = 8},
CmdPart {arg = "https://user:pass@eg.com/foo", partId = 12},
CmdPart {arg = "--extra-index-url", partId = 14},
CmdPart {arg = "https://user:pass@ex-eg.io/foo", partId = 18},
CmdPart {arg = "foobar==1.0.0", partId = 20}
]
```
Extra-index-url id is 14 and it tries to remove item with id 16, but that id doesn't exsist. 

### How I did it
Search for the smallest argument id which is greater than the flag id. 

### How to verify it
Tests mentioned in issue was added in Spec file.


lorenzo:
ah, great find! Thanks so much :)

PR #474, opened 2020-10-06
add proxy attribute | pip

dariodsa:
### What I did
I add proxy flag in pip command array. 

```
--proxy <proxy>             Specify a proxy in the form [user:passwd@]proxy.server:port
```


lorenzo:
Thanks again 💯

Issue #475, opened 2020-10-13
Non-deterministic failures (exit code 127)


omh1280:
- [x] This is a bug report
- [ ] This is a feature request
- [x] I searched existing issues before opening this one

### Expected behavior

Hadolint will exit with code `0` if run against a valid Dockerfile.

### Actual behavior

Hadolint rarely exits with code `127` on a Dockerfile that should pass.

This is reproducible on Mac and Linux and can be tested with the below dockerfile. Since this bug seems non-deterministic, the value of `1000` is chosen to sufficiently increase the probability that the run will fail.

We use `parallel` to speed up running hadolint 1000 times. This has been seen on CI/CD infrequently, which is why this issue is being raised.

### Steps to reproduce the behavior

**Bash**
```bash
$ mkdir hadolint-stress && cd hadolint-stress
$ touch Dockerfile   # copy dockerfile below into here
$ docker build -t hadolint-stress .
$ docker run hadolint-stress   # should exit with 127 most of the time
```

**Dockerfile**

```dockerfile
FROM ubuntu:focal

RUN apt-get update && apt-get install --no-install-recommends -y \
  wget=1.20.3-1ubuntu1 \
  ca-certificates=20190110ubuntu1.1 \
  parallel=20161222-1.1 \
  && rm -rf /var/lib/apt/lists
RUN wget -q https://github.com/hadolint/hadolint/releases/download/v1.18.0/hadolint-Linux-x86_64 -O hadolint \
  && chmod +x hadolint

COPY ./Dockerfile ./Dockerfile

# hadolint ignore=SC2039
CMD [ "/bin/bash", "-c", "parallel -N0 --halt now,fail=1 ./hadolint ./Dockerfile ::: {1..1000}; echo $?" ]
```

If the value of `1000` is reduced to `100`, a majority of the runs pass:

```bash
$ docker run hadolint-stress
0
$ docker run hadolint-stress
0
$ docker run hadolint-stress
0
$ docker run hadolint-stress
0
$ docker run hadolint-stress
0
$ docker run hadolint-stress
parallel: This job failed:
./hadolint ./Dockerfile
127
```

**Output of `hadolint --version` or
  `docker run --rm hadolint/hadolint hadolint --version`:**  Haskell Dockerfile Linter v1.18.0-0-g76eee5c (seen in 1.17 as well)

**Additional environment details (OS, stack version, etc.)** MacOS & Linux

m-ildefons:
Hi @omh1280 ,
[...]

PR #479, opened 2020-10-19
Run the linter in parallel when multiple files supplied

lorenzo:
### What I did

A use case surfaced by #475 is that some people lint lots of files, and try to make things faster by running them in parallel using bash.

Running things in parallel in Haskell is trivial, so I went ahead and added this feature. After some superficial testing, found out that the best performance is reached at maximum 4 CPUs at a time. That's why I capped the `-N` to 4.

Whoever wants to experiment with more CPUs may invoke `hadolint` like this:

```
hadolint +RTS -N[num] -RTS Dockerfile1 Dockerfile2 Dockerfile3...
```

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -28,6 +28,8 @@ dependencies:
   - language-docker >=9.1.2 && < 10
 library:
   source-dirs: src
+  generated-other-modules:
+    - Paths_hadolint
   dependencies:
     - &bytestring bytestring >=0.10
     - &split split >=0.2
@@ -41,6 +43,7 @@ library:
     - HsYAML
     - filepath
     - directory >= 1.3.0
+    - async
 executables:
   hadolint:
     main: Main.hs
@@ -51,6 +54,11 @@ executables:
       - gitrev >=1.3.1
       - text
       - containers
+    ghc-options:
+      - -O2
+      - -threaded
+      - -rtsopts
+      - -with-rtsopts=-N
     when:
       # OS X does not support static build https://developer.apple.com/library/content/qa/qa1118
       - condition: "flag(static) && !(os(osx))"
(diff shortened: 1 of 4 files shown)


zemanlx:
Does this mean that hadolint is now running checks in parallel or Dockerfiles in parallel? What that `RTS` means? I am not a huge fan of parallel running as it usually causes more issues than benefits, but I will hope that this is not the case. I cannot see where `-N` is capped to 4. Can you please show me? If we want to keep this feature, can you please add it to docs and maybe add check in CI that is testing this feature?

lorenzo:
* Dockefiles are processed parallel, not rules.
* `RTS` stands for `Run Time System` it's a way to pass flags to the haskell RTS to change things like the garbage collector and number of CPUs to use
* It seems like I ended up capping it at 5 (https://github.com/hadolint/hadolint/pull/479/files#diff-e0ee4e21f8811c1171864cc68ea4005347b1b0ca70626026f251bf4111c2aa6eR58) The extra CPU I measured to be good for garbage collection efficiency.


I'll add some docs and tests for it, although not sure how to best test that something ran in parallel, ideas? @zemanlx

zemanlx:
Thank you! Why do we need to cap it? Should not it just depend on the number of CPUs/cores that the machine running it has?

For CI I think we can repurpose [integration test](https://github.com/hadolint/hadolint/blob/master/integration_test.sh#L106) which has tons of Dockefiles.

lorenzo:
Many things, when run in parallel end up suffering from diminishing returns. When the tasks to run are not hard enough, or when each task needs to fight for resources, then the wait time ends up taking more CPU time than the actual tasks. The right combination is usually found via experimentation. This was the number with the lowest runtime after many tests processing 1110 files.

While using the 12 CPUS in my machine was possible, and they were all busy, the runtime was actually slightly worse.

Issue #467, opened 2020-10-05
Python requirements.txt


Dev-Crea:
- [] This is a bug report
- [x] This is a feature request
- [x] I searched existing issues before opening this one

### Expected behavior

No warning when install package python with option `--requirement` (or `-r`).
I use `requirements.txt` for install python dependencies with a precise a version.

### Actual behavior

Have a linter error :
```bash
docker run --rm -i hadolint/hadolint:latest-alpine < Dockerfile
/dev/stdin:20 DL3013 Pin versions in pip. Instead of `pip install <package>` use `pip install <package>==<version>`
```

### Steps to reproduce the behavior

**Output `docker run --rm hadolint/hadolint hadolint --version`:**
```bash
Haskell Dockerfile Linter v1.18.0-15-g6664056
```

**Simply launch linter with docker :**
```bash
docker run --rm -i hadolint/hadolint:latest-alpine < Dockerfile
```

**Dockerfile** 

```Dockerfile
FROM ubuntu:20.04

ARG DEBIAN_FRONTEND=noninteractive

ENV AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
ENV AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}
ENV AWS_REGION=${AWS_REGION}
ENV TZ=Europe/Paris

WORKDIR /app

COPY requirements.txt .

SHELL ["/bin/bash", "-o", "pipefail", "-c"]

# hadolint ignore=DL3013
RUN apt-get update \
  && apt-get install -y --no-install-recommends \
    python3-dev=3.8.2-0ubuntu2 \
    python3-pip=20.0.2-5ubuntu1 \
    curl=7.68.0-1ubuntu2.2 \
  && ln -s /usr/bin/pip3 /usr/bin/pip \
  && curl -s https://packagecloud.io/install/repositories/dazzltv/pypkg/script.python.sh | bash \
  && pip install --upgrade pip \
  && pip3 install --requirement requirements.txt --target . \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/* \
  && apt-get remove -y \
    curl=7.68.0-1ubuntu2.2

CMD ["python3", "./task.py"]
```

PR #484, opened 2020-10-28
dl3038 pip --no-cache-dir rule

### LINKED ISSUE #455: Raise an error if `pip install` doesn't have `--no-cache-dir` option provided
- [ ] This is a bug report

### Expected behavior

Let's build an image where the `pip` has the enabled cache

```shell
docker build -t pip-with-cache - <<EOF
FROM python:3.8-alpine
RUN python -m pip install --upgrade pip && python -m pip install flask
EOF
```

Let's build another image where `pip` has the disabled cache

```shell
 docker build -t pip-no-cache - <<EOF
FROM python:3.8-alpine
RUN python -m pip install --no-cache-dir --upgrade pip && python -m pip install --no-cache-dir flask
EOF
```

Let's compare images

```shell
docker image ls --filter reference="pip*"
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
pip-no-cache        latest              64873c9565ed        8 seconds ago       94.7MB
pip-with-cache      latest              205c4c08f43f        51 seconds ago      96.9MB
```

As the output shows, the image with disabled cache has a smaller size. And there is only one package installed. That's why I propose to add a new rule that will require adding `--no-cache-dir` after `pip` install.

### Actual behavior

There is no error.

jhmcstanton:
Includes tests and reference to non-existing wiki page.

Resolves #455

### What I did

Added a new rule for `pip --no-cache-dir` usage

### How I did it

Added the rule to `Rules.hs` and its exported `rules` list. This also factored out some code from the `pipVersionPinned` rule so that both rules can share the `isPipInstall` logic.

### How to verify it

The provided tests and the Dockerfiles referenced in #455 and  #467

If this change is accepted I can create the related wiki entry, just let me know if that is needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,6 +211,7 @@ Please [create an issue][] if you have an idea for a good rule.
 | [DL3039](https://github.com/hadolint/hadolint/wiki/DL3039)   | Do not use `dnf update`                                                                                                                             |
 | [DL3040](https://github.com/hadolint/hadolint/wiki/DL3040)   | `dnf clean all` missing after dnf command.                                                                                                          |
 | [DL3041](https://github.com/hadolint/hadolint/wiki/DL3041)   | Specify version with `dnf install -y <package>-<version>`                                                                                           |
+| [DL3042](https://github.com/hadolint/hadolint/wiki/DL3038) | Avoid cache directory with `pip install --no-cache-dir <package>`.                                                                                  |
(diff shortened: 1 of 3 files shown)


lorenzo:
sorry for the dealy reviewing this. Would you mind solving the conflicts @jhmcstanton ?

jhmcstanton:
Updated, no worries about the delay.

PR #487, opened 2020-11-04
Yum lint - Add negative tests

m-ildefons:
### What I did

Add tests making sure `yum` lint rules are not overzealous.

### How I did it

- Test against correct use of `yum` in each case
- Test against no use of `yum` at all in each case

Both times the rule should not trigger.

### How to verify it

Run `stack test`. All tests should pass.


zemanlx:
Thank you again :tada:

PR #491, opened 2020-11-14
Feature: Extra Rules & Healthcheck Rule.

### LINKED ISSUE #360: Check healthcheck presence

### Expected behavior

When there is no HEALTHCHECK, the linter should throw an error (DL 3029)

m-ildefons:
Fixes #360

### What I did
* Created a new category of rules: `extra`, which are opt-in.
  These rules can be enabled with the commandline option `--extra=DLxxxx` or via config file:
```
extra:
  - DLxxxx
```

* Add rule DL4007 checking for `HEALTHCHECK` instructions.
* Add mandatory rule DL4008 checking for multiple `HEALTHCHECK` instructions

### How I did it

The rule DL4007 is a bit of a hack, since it essentially matches on the `FROM` instruction in cases where no line with HEALTHCHECK was found in the dockerfile. I think this is acceptable though, since a Dockerfile without a `FROM` statement is moot anyways. It is opt-in, because in some cases (e.g. Kubernetes) the health check should be omitted.

The machanism for the extra rules was added analogous to ignored rules: The extra rules are added to the list of rules to check. By default, no rules from that list is added, but every named rule, that is also a member of the extra rule list is added to the list of rules to check.

Note, that ignore takes precedence over extra. This is important in the following usecase. When you set up a CI environment based on Hadolint, you can include a set of extra rules for the whole environment based on a config file, but then you can still ignore particular rules on a per-Dockerfile or per-buildjob basis using any existing ignore mechanism.

The name `extra` for these rules is a compromise. I should have named it supplemental or auxiliary, but I wanted to have the brevity for convenience on the commandline.

### How to verify it

This triggers no error:
```Dockerfile
FROM debian:buster
```
But is does trigger DL4007 when explicitly enabled like so: `hadolint --extra=DL4007 Dockerfile` or via config file.

This triggers DL4008:
```Dockerfile
FROM debian:buster
HEALTHCHECK CMD a
HEALTHCHECK CMD b
```

In addition, the two rules and the config-parser each have their unit-tests, so a `stack test` should still pass.


lorenzo:
I really like the idea. I'm wondering if we should make this more generic and we provide a way to filter rules based on their severity or another attribute, that was we avoid having two lists of rules and open the door for filtering out rules by severity.

m-ildefons:
That's an interesting idea, I haven't thought of. Can you please elaborate on how you would like this more generic filtering to look like? In my mind, this could look something like so:

```yaml
# .hadolint.yaml
severity:
  - DL3008: warning
  - DL4007: warning
  - DL3013: ignore
```

with the equivalent commandline being:
```shell
$ hadolint --warning=DL4007 --warning=DL3008 --ignore=DL3013 Dockerfile
```

This does however break compatibility with current `.hadolint.yaml` files, which this PR does not do.
As Hadolint is often used as part of a CI/CD pipeline, I would reconsider where we want to go with the more generic filtering. My question is what is the use case here?
In my mind at least, it is not terribly useful in a CI/CD pipeline and has only limited use for a developer on the command line, because in these two cases, only output/no-output is really interesting. Only when used as a linter plugin for an editor, customizable severities provide a noticeable benefit. Or am I missing something here?

That said, I can totally see the benefit in not having two lists of rules with one silently taking precedence over the other.

lorenzo:
@m-ildefons  This feature request has come up many times in the past. The use case is that some companies enforce more strictly certain rules and others would like to get informed about a violation but not break the pipeline when it happens.

For instance, I may be interested in knowing that my images are not pinning dependencies, but would not like the build to break because we are in a transition period.

Changing the config file is not a concern as we can make it backwards compatible.

m-ildefons:
@lorenzo I've read your comment above again an realized, that that is really several different features:
1) customizable severities of rules
2) filtering out rules based on severity
which are again different from what I need for the healthcheck rule, which is default-off rules that a user would have to opt in.

Customizable severities of rules is related to this PR, since the healthcheck rule needs customizable severity to be opt-in. The filtering part is a different feature.

The implementation is a bit tricky, since the severity is simply Shellchecks severity, which only supports four levels:
`error,` `warning`, `info`, `style`. Ideally, we would have another level `ignore`, so a rule could be ignored by default and then the level of that rule could be raised on request of the user.

Implementing the mechanism to modify the severity of a specific rule on request of a user is easy, but still bound to the severities from Shellcheck.
That means that two lists of rules is the easy way to implement opt-in rules. The other way would be to change the severity of a RuleCheck to a `Maybe Severity`, which would require quite extensive code modifications, but would allow rules to have a severity of `Nothing` by default which could then be raised to `Just X` on request of the user.
This is a very intrusive change and before I sink a day or two into trying to make it I'd like to hear your opinion on whether or not it would be worth it.

For now, I've implemented customizable severity here: https://github.com/m-ildefons/hadolint/tree/custom-severity but like I said, this does not yet support opt-in rules, which is needed here. Maybe you have a better idea how to go about this?

lorenzo:
@m-ildefons What I was thinking is that all rules are always checked, but the `extra` rules have a default severity of `Notice` or something like that. That way you get notified about the rule, but the build does not break. 

How does that sound?

m-ildefons:
Well, the build breaks whenever Hadolint exits with status not equal to 0, which happens if any rule of any severity matches. What I want here are rules that are not even checked, if not explicitly requested.

Take the healthcheck rule here for example. Unless you are in the special situation, that you actually care about this, you probably don't want to even see the message. IMO, this is the preferred default for this rule, because the healthcheck feature of docker is pretty niche.
I'll admit, that it might then be a good idea to add a `--brutal`-mode to just opt into all rules.

I dug a bit into it and found that is isn't hard to make the severity a `Maybe`. This can be leveraged to have rules that are not checked at all, except when their severity is manually increased, which is exactly what I want.
I've updated my test branch and will probably rebase this PR onto that, since it is much cleaner.

m-ildefons:
Hi @lorenzo ,

I've revised the design for the optional rules. Now it is much more in line with what I understood to be your vision here. You can specify the desired severity of any rule like so:
```yaml
error:
  - DL3008

warning:
  - DL4001

info:
  - DL3017

style:
  - DL3009

ignore:
  - DL3001
```
or by using the command line:
```
$ hadolint --error=DL3008 --warning=DL4001 --info=DL3017 --style=DL3009 --ignore=DL3001 Dockerfile
```
This is in line with the previously available severities and avoids the use of multiple rule lists.
In addition to that a rule may be implemented with a severity of `Nothing`. It will then not be evaluated, except when the user explicitly enables it by overriding its severity.
What do you think? Is this closer to your vision of this feature?

app/Main.hs
```diff
@@ -104,7 +139,12 @@ parseOptions =
 
     files = many (argument str (metavar "DOCKERFILE..." <> action "file"))
 
-    lintOptions = Hadolint.LintOptions <$> ignoreList <*> parseRulesConfig
+    lintOptions = Hadolint.LintOptions <$> errorList
```
lorenzo, line 142: I would nest this new stuff below an `override` key:

```yaml
override:
    error:
      - DL3002
```

src/Hadolint/Formatter/Codeclimate.hs
```diff
@@ -83,14 +82,6 @@ checkToIssue RuleCheck {..} =
       impact = severityText (severity metadata)
     }
 
-severityText :: Severity -> String
```
lorenzo, line 86: This code is required by `codeclimate` as those are the values that it can read

src/Hadolint/Formatter/Format.hs
```diff
@@ -46,13 +44,14 @@ toResult res =
...
   case s of
-    ErrorC -> "error"
-    WarningC -> "warning"
-    InfoC -> "info"
-    StyleC -> "style"
+    Just ErrorC -> "error"
+    Just WarningC -> "warning"
+    Just InfoC -> "info"
+    Just StyleC -> "style"
+    Nothing -> "ignore"
```
lorenzo, line 54: Will this be printed somewhere? It would be weird to see `ignore`d stuff

src/Hadolint/Rules.hs
```diff
@@ -860,11 +861,35 @@ usePipefail = instructionRuleState code severity message check False
...
+hasHealthcheck dockerfile = instructionRuleState code severity message check Nothing dockerfile
+  where
+    code = "DL4007"
+    severity = Nothing
+    message = "No `HEALTHCHECK` instruction"
+    check _ _ From {}
+      | null (allHealthchecks dockerfile) = withState Nothing False
+      | otherwise = withState Nothing True
+    check st _ _ = withState st True
+    allHealthchecks df = [(l, h) | (l, Healthcheck h) <- instr]
```
lorenzo, line 874: I'm not sure this is correctly captures the intent of the rule. I think the rule should be like this:

* For each `FROM` remember whether it had a health check or not
* Return `False` if the last `FROM` did not have a health check
  * This mean, if the last `FROM` inherits from a previous stage, and any of the previous has a health check, the you can return `True`

I suggest adding tests for those cases too

lorenzo (changes requested):
This is looking great!

It captures exactly what I meant before. I'm still split between whether a `Maybe Severity` is the right thing or not. Was wondering if it would be better to have our own `Severity` and not use the one from ShellCheck. That would allow us to have an `Ignore` severity.

I left a few comments on some implementation issues that I spotted.

app/Main.hs
m-ildefons: Agreed. This sounds reasonable.

src/Hadolint/Formatter/Codeclimate.hs
m-ildefons: Oops, how embarrassing. I mistook this to be the same as the other `severityText`

src/Hadolint/Formatter/Format.hs
m-ildefons: No. Rules with a severity of `Nothing` will not produce any output. But it would be weird to have a non-exhaustive match here and `ignore` is the most sane return value I could come up with.

src/Hadolint/Rules.hs
m-ildefons: This captures exactly your desired behaviour, maybe not in an obvious way:
If there is a `HEALTHCHECK` at any point in the dockerfile, this rule evaluates to `True` for all lines. If there is no `HEALTHCHECK` at all, it evaluates to `True` for all lines that are not `FROM`s, but `False` for all lines that are. 
As a result, none of the following dockerfiles produce a warning:

```Dockerfile
FROM debian:buster as build
HEALTHCHECK CMD a
FROM debian:buster
```

```Dockerfile
FROM debian:buster as build
FROM debian:buster
HEALTHCHECK CMD a
```

```Dockerfile
FROM debian:buster as build
HEALTHCHECK CMD b
FROM debian:buster
HEALTHCHECK CMD a
```

I'll add tests for that though.

test/Spec.hs
```diff
@@ -971,6 +967,61 @@ main =
...
+              ]
+        in ruleCatchesNot hasHealthcheck $ Text.unlines dockerFile
+
+      it "multistage with healthcheck in early stage only passes" $
+        let dockerFile =
+              [ "FROM debian:buster as stage1",
+                "HEALTHCHECK CMD check-1",
+                "FROM debian:buster"
+              ]
+        in ruleCatchesNot hasHealthcheck $ Text.unlines dockerFile
```
lorenzo, line 1008: I don't understand how this example passes the rule? The last `FROM` has no `HEALTCHECK` and is not inheriting from the first `FROM`

test/Spec.hs
```diff
@@ -971,6 +967,61 @@ main =
...
+              ]
+        in ruleCatches multipleHealthcheck $ Text.unlines dockerFile
+
+      it "not too many healthchecks" $
+        let dockerFile =
+              [ "FROM busybox",
+                "HEALTHCHECK CMD ls /1"
+              ]
+        in ruleCatchesNot multipleHealthcheck $ Text.unlines dockerFile
+
```
lorenzo, line 1024: I'm missing this case:

```dockerfile
FROM debian:buster as stage1
HEALTCHECK foo

FROM stage1
```

test/Spec.hs
m-ildefons: That's not at all what this rule was supposed to check.
DL4007 was only about whether or not you have defined a `HEALTHCHECK` somewhere in you Dockerfile.

I have grown a few grey hairs over a revised version that has more sophisticated behaviour with multistage builds since.

Now this case and the case below are handled correctly and tested for - as well as the more complicated cases of multistage builds with and without consecutive build stages.

test/Spec.hs
m-ildefons: Line 1011 ff. ?

m-ildefons:
Hi @lorenzo , I'd like to move this on, but it seems you are not yet happy with with PR.

lorenzo:
@m-ildefons sorry, it is not that I'm unhappy with the PR, quite the opposite. I just took a vacation from maintaining opensource projects. There are certain changes that I would like to make to the code.

Since I made you wait this long, I'll try to add them myself

m-ildefons:
I took the liberty to rebase this onto master. If you like, I'll see if I can pull this apart in smaller self contained PRs so that it is not such a huge chunk of change all at once.

lorenzo:
@m-ildefons I was trying alternatives to using a `Maybe Severity` to represent this, as I still think that it complicates the code more than I would like to

m-ildefons:
I'll abandon this and rebase the two `HEALTHCHECK` rules on top of #543

PR #493, opened 2020-11-15
feature: `pip install` from local archive no longer emits DL3013 #485

### LINKED ISSUE #485: DL3018 Should be surpressed when installing from archive
- [X] This is a feature request
- [X] I searched existing issues before opening this one

### Expected behavior

Pip supports installing from local archive files. In a docker file, it would look like:

``RUN pip install mypkg.whl``

hadolint emits a DL3018 when it encounters this. This is a common construct used by developers to install their code into a container. Having to specify (and repeatedly update) the version could be challenging during development.

If you just suppress DL3018, that would apply to all the packages being installed from the same command, which means long RUN commands (to reduce layer count) would loose this valuable advice.

m-ildefons:
Fixes #485

### What I did
- Exclude local archives from emitting DL3013.
- Test that local archives no longer emit DL3013

Emitting DL3013 for local packages is pointless, since the developer of the dockerfile controls what that local package is anyways and therefore also controls the version of the software within.

### How to verify it
This Dockerfile no longer triggers DL3013:
```Dockerfile
RUN pip install mypkg.whl
RUN pip install mypkg.tar.gz
```

This Dockerfile still triggers DL3013:
```Dockerfile
RUN pip install numpy
```

PR also comes with unit tests.

### DIFF
--- a/src/Hadolint/Rules.hs
+++ b/src/Hadolint/Rules.hs
@@ -643,10 +643,12 @@ pipVersionPinned = instructionRule code severity message check
               "upgrade-strategy"
             ]
             cmd
-    versionFixed package = hasVersionSymbol package || isVersionedGit package
+    versionFixed package = hasVersionSymbol package || isVersionedGit package || isLocalPackage package
     isVersionedGit package = "git+http" `Text.isInfixOf` package && "@" `Text.isInfixOf` package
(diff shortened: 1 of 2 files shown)


lorenzo:
awesome, thanks!

PR #496, opened 2020-11-19
Replacing travis with github actions

lorenzo:
### What I did

Replaced Travis and Appveyor with Github actions. The ecosystem seems to be converging around it now

Some highlights:

* It was a lot easier to build with `cabal` instead of `stack`, especially because the cache got corrupted frequently in macos.
* Hlint suggestions are now added as comments in the pull request
* Artifact names will change
* Not using `--brutal` for upx

## Hlint comments
Screenshot 2020-11-20 at 12 56 17

## New artifacts name
Screenshot 2020-11-20 at 19 13 17


lorenzo:
@zemanlx I think this is ready to review. Can you take a look?

zemanlx:
@lorenzo That is some serious work done! I was thinking about this for some time, but I am glad you have actually done it. Thank you!

I have a few questions:
- Does this mean that we are leaving `stack` for `cabal`? 
- If not, do you plan to test against the latest `lts` or only against pinned version?
- Do you think it would be possible to keep names of release binaries? I believe that a lot of people have these hardcoded (including me) in their projects and upgrading by switching version number in a variable and this will be breaking change.

lorenzo:
@zemanlx Thanks!

For the time being, I'm thinking of only using cabal for GitHub actions but continue to support stack for local development, which I prefer. I will periodically try again to see if we can use `stack` in GitHub actions without stumbling upon the same caching bug. If enough time passes and we cannot fix it, then I'd recommed entirely moving to cabal.

For the time we are not testing with stack, we are just going to use the pinned versions. I can try in another iteration to make cabal use the `lts` selection of packages.


I'll also try how difficult it would be to maintain the same file names

lorenzo:
@zemanlx ok, renaming the binaries was easier than I thought. Let me know your thoughts

zemanlx:
@lorenzo Sounds good enough to me. Thanks to making binary names backwards compatible!

I will probably miss a bit that CI is now working a bit different than local development, seeing that new `lts` is breaking the build was quite handy and was forcing us to keep it up to date, but it is not a biggie.

zemanlx (approved):

PR #498, opened 2020-11-20
Fix link to DL3042

pquentin:
I tried to understand rule DL3042 but the link in the README leads to another one. This fixes the link, but the wiki page does not exist yet and should be created.


pquentin:
I actually went ahead and created https://github.com/hadolint/hadolint/wiki/DL3042

lorenzo:
thanks!

PR #502, opened 2020-11-27
rules: multiple `COPY` to same location

### LINKED ISSUE #439: Linting of multiple instance of COPY to same location
- [] This is a bug report

### Expected behavior
Show linting error while doing to avoid creating unnecessary layers
COPY ./file1 ./temp
COPY ./file2  ./temp
COPY ./file3 ./temp

### Actual behavior
No linting error found 

```bash
(paste your output here)
```

**Dockerfile** (if relevant)

```Dockerfile
(paste content of your Dockerfile)
```

m-ildefons:
### What I did

- Add rule to check against multiple `COPY` instructions to the same
location.
- Add tests and README entry for said rule.

The rule checks that no two `COPY` instructions have the same
destination path either within the same build stage, or within the base
build stage.
It tracks destinations across multiple build stages.

Closes #439 
### How I did it

The rule tracks all `COPY` instruction in a list of tuples that looks like this:

```haskell
(currentBuildStageIdentifier :: Data.Text, [(destination :: Data.Text, BuildStageIdentifier :: Data.Text)]
```
currentBuildStageIdentifier is the image alias from the last `FROM` instruction or if that does not exists, simply the image name from the last `FROM` instruction.

When it enters a new build stage it copies all destinations whose BuildStageIdentifier matches the current base image and assigns them the current image alias as BuildStageIdentifier, thus inheriting all `COPY` destinations from the previous buid stage.

### How to verify it
There are unit tests.
```Dockerfile
FROM baseimage as base1
COPY foo /asdf  # ok
COPY bar /asdf  # not ok
FROM baseimage as base2
COPY baz /asdf  # ok
FROM base1 as base3
COPY bla /buzz  # ok
COPY bum /asdf  # not ok
FROM base1 as base4
# noop
FROM base4
COPY qwer /asdf  # not ok
```


zemanlx:
Thank you @m-ildefons! I have a question regarding behaviour. Why is overwriting file between stages not OK?

m-ildefons:
HI @zemanlx ,
I decided to flag it, because it leads to unnecessarily obfuscated build files.

The only possibly slightly reasonable docker build I could come up with where it might be desirable to have two `COPY` instructions to the same location, is something like this:

```Dockerfile
FROM baseimage as builder
COPY build.sh /start.sh
RUN /start.sh

FROM builder
COPY run.sh /start.sh
CMD ["/start.sh"]
```

Note, how the files already have two different names prior to being copied and how this really is not any better than:
```Dockerfile
FROM baseimage as builder
COPY build.sh /build.sh
RUN /build.sh && rm /build.sh

FROM builder
COPY run.sh /run.sh
CMD ["/run.sh"]
```

I will admit that it does not make a difference in layer count or size, but the separation in a multistage build is kind of pointless here as well. Anyways, my point is that when `COPY`ing to the same location, it is far more likely to do a mistake than it is to be clever **and** understandable about it.

zemanlx:
If/when we have different levels of importance I would consider this one as INFO as I can think of legitimate use and I do not think we should error if we are not sure that we are pointing to the real issue. But I think it may cause confusion and therefore it is worth to let a user know.

My valid use case is to have one base image with the default configuration which is then reused for some specific scenarios with some specific configs which overwrite the default one. What do you think @m-ildefons?

m-ildefons:
I did not know that you could produce multiple images off of a single dockerfile, but I guess if untagged images count, you can.

Anyways, considering that, yeah INFO severity is a better choice.

m-ildefons:
@zemanlx , I've removed the tracking across different build stages and reimplemented the state as `Set` of destinations rather than a list. I've also changed the message to more accurately match the new behavior of the rule.

I still think that overwriting files from earlier build stages is a bad practice and should be warned about, but maybe not in this rule.

lorenzo:
This is also good, just need to solve the conflicts again :P

test/Spec.hs
```diff
@@ -1217,6 +1217,38 @@ main =
               ]
          in ruleCatchesNot copyFromAnother $ Text.unlines dockerFile
     --
+    describe "multiple COPY to same destination" $ do
+      it "not ok with multiple COPY to same destination - single stage" $
+        let dockerFile =
+              [ "FROM baseimage",
+                "COPY foo /asdf",
```
lorenzo, line 1224: There is a very common pattern that I use often in my docker files:

```
FROM node
COPY package.lock package.json /app/
RUN npm install
COPY src /app
```

Which is a valid pattern as it allows docker to cache more efficiently. I think this rule will flag that incorrectly

m-ildefons: Good catch. I've played around with different ways to copy directories into docker images and I'm starting to wonder, if I have attempted the impossible with this rule. You see, the destination can be pretty much anything and there is no way to distinguish between files and directories without inspecting the filesystem within the image.

I'm out of ideas here. Maybe this PR should be abandoned.

m-ildefons:
I haven't come up with a way to avoid false positives and I doubt, that it can even be done. I'll abandon this PR.

PR #517, opened 2021-01-20
DL3005: Fix missing dist-upgrade #514

### LINKED ISSUE #514: DL3005 Do not use apt-get upgrade or dist-upgrade - Doesn't care about dist-upgrade
- [x] This is a bug report

### Expected behavior

`DL3005` should fail if there is `dist-upgrade`

### Actual behavior

Doesn't fail

```Dockerfile
FROM debian:buster
RUN dist-upgrade

```

https://github.com/hadolint/hadolint/blob/master/src/Hadolint/Rules.hs#L429

m-ildefons:
- Check for `apt-get dist-upgrade` in addition to the existing
  check for `apt-get upgrade`

Performing `apt-get dist-upgrade` is to be considered bad practice for
the same reasons that `apt-get upgrade` would be considered bad
practice. Perhaps it should be considered worse in cases where an
appropriate base image is available.

### What I did
Fixes #514
### How I did it
Add "dist-upgrade" to the list of command line arguments for apt-get to be checked for.
### How to verify it
This should trigger `DL3005`:
```Dockerfile
RUN apt-get dist-upgrade
```
Unit tests are included.

### DIFF
--- a/src/Hadolint/Rules.hs
+++ b/src/Hadolint/Rules.hs
@@ -428,7 +428,7 @@ noAptGetUpgrade = instructionRule code severity message check
     severity = ErrorC
     message = "Do not use apt-get upgrade or dist-upgrade"
     check (Run (RunArgs args _)) =
-      argumentsRule (Shell.noCommands (Shell.cmdHasArgs "apt-get" ["upgrade"])) args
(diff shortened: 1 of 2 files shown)


zemanlx (approved):
Thank you @m-ildefons!

06kellyjac:
Thanks

PR #523, opened 2021-01-28
add pre-commit hook configuration

wwuck:
### What I did

Add a yaml configuration file to allow running hadolint as a git pre-commit hook with [pre-commit](https://pre-commit.com/).  Once set, the yaml hook configuration file doesn't need to change.  The hook allows either using a local hadolint binary or using the official hadolint Docker image.

### How to verify it

Install pre-commit from the docs at https://pre-commit.com/ and install the hooks with `pre-commit install` in the git repository.  Add a `.pre-commit-config.yaml` file following the docs in this pull request, and then attempt to commit some staged changes to any Dockerfile into git.  `pre-commit` only runs on changed files by default so it won't run on any Dockerfiles that are not included in the current commit.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,13 @@
+---
+- id: hadolint-docker
+  name: Lint Dockerfiles
+  description: Runs hadolint Docker image to lint Dockerfiles
+  language: docker_image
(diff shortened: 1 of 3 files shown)


lorenzo:
thanks, this is great!

Issue #512, opened 2021-01-08
Smell for ONBUILD Sub-instructions


k----n:
- [ ] This is a bug report
- [x] This is a feature request
- [x] I searched existing issues before opening this one

As noted in the Dockerfile docs (https://docs.docker.com/engine/reference/builder/#onbuild):
-  Chaining ONBUILD instructions using ONBUILD ONBUILD isn’t allowed.
- The ONBUILD instruction may not trigger FROM or MAINTAINER instructions.

PR #526, opened 2021-01-31
rules: Check against illegal instructions within `ONBUILD` #512

m-ildefons:
- Add rule to check agains illegal instructions within `ONBUILD`
  instructions
- Modify rule parsing to do a pass with and a pass without automatically
  expanded `ONBUILD` instructions
- Merge results of the two passes such that all rules appear once only
- Add tests for new rule

I'd like to point out two things for discussion regarding this change:
- In addition to the three cases checked here, IMO not all other combination that are legal by this rule make a lot of sense.
  For example `ONBUILD EXPOSE` or `ONBUILD LABEL` do not appear very useful to me - maybe we should warn on them?
  The requested three cases are to be avoided as per official docker documentation, so I think there is no need to discuss the necessity of linting for them.
- Checking the `ONBUILD` instructions themselves requires a second pass through all instructions. This is not too bad IMO, but
  maybe there is a better way. I tried to use pattern matching to only check the `ONBUILD` instructions twice, but failed to come up with something due to my limited Haskell knowledge.
  The reason the `ONBUILD` instructions absolutely need to be checked twice is that `hadolint` must also lint whatever is inside the `ONBUILD` instruction. Maybe there is a better way that I just can't see right now?

### How to check
This dockerfile should trigger error `DL3043` on all three lines:
```Dockerfile
ONBUILD FROM debian:buster
ONBUILD MAINTAINER "Some Guy"
ONBUILD ONBUILD RUN something
```

This dockerfile does not trigger `DL3043`:
```Dockerfile
FROM debian:buster
MAINTAINER "Some Guy"
ONBUILD RUN something
```

### DIFF
--- a/src/Hadolint/Rules.hs
+++ b/src/Hadolint/Rules.hs
@@ -91,10 +91,14 @@ mapInstructions f initialState dockerfile =
   let (_, results) = mapAccumL applyRule initialState dockerfile
    in concat results
   where
-    applyRule state (InstructionPos (OnBuild i) source linenumber) =
-      applyWithState state source linenumber i -- All rules applying to instructions also apply to ONBUILD,
+    applyRule state (InstructionPos onbuild@(OnBuild i) source linenumber) =
+      -- All rules applying to instructions also apply to ONBUILD,
       -- so we unwrap the OnBuild constructor and check directly the inner
-      -- instruction
+      -- instruction. Then we also check the instruction itself and append the
+      -- result to avoid losing out on detecting problems with `ONBUILD`
+      let (innerState, innerResults) = applyWithState state source linenumber i
+          (finalState, outerResults) = applyWithState innerState source linenumber onbuild
+       in (finalState, innerResults <> outerResults)
     applyRule state (InstructionPos i source linenumber) =
       applyWithState state source linenumber i -- Otherwise, normal instructions are not unwrapped
     applyWithState state source linenumber instruction =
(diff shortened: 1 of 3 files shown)


src/Hadolint/Rules.hs
```diff
@@ -89,12 +89,19 @@ type Rule = ParsedFile -> [RuleCheck]
 mapInstructions :: CheckerWithState state -> state -> Rule
 mapInstructions f initialState dockerfile =
   let (_, results) = mapAccumL applyRule initialState dockerfile
-   in concat results
+      (_, nestedResults) = mapAccumL nestedApplyRule initialState dockerfile
+   in concat $ results ++ [r | r <- nestedResults, r `notElem` results]
   where
-    applyRule state (InstructionPos (OnBuild i) source linenumber) =
```
lorenzo, line 94: I believe this could be written without the second pass and the merging of the rules by calling `applyWithState` twice on the `OnBuild` case like this:

```
    applyRule state (InstructionPos onbuild@(OnBuild i) source linenumber) =
      -- All rules applying to instructions also apply to ONBUILD,
      -- so we unwrap the OnBuild constructor and check directly the inner
      -- instruction
      let (innerState, innerResults) = applyWithState state source linenumber i
          (finalState, outerResults) = applyWithState innerState source linenumber onbuild
       in (finalState, innerResults <> outerResults)
```

m-ildefons: This is great! This is exactly what I had in mind. Thank you for showing me :-)

lorenzo:
awesome work, thanks!

PR #528, opened 2021-02-01
Add '--no-fail' option

### LINKED ISSUE #391: Suppress Exit Code

antechrestos:
Each formatter shall exit non 0 code when any error occurs unless '--no-fail' option specified. If so, a 0 status code is returned even if any rule is violated

Closes #391

### What I did

Make program return non zero code if failure occurs, whenever the formatter. If `--no-fail` is specified , the programm returns `0` status code, whatever rules are violated or not.

### How I did it

- remove treatment of exit in lint and put it in main
- added option
- print result and analyse the exit method that shall be called

### How to verify it

- launch binary with invalid `Dockerfile` with `-f gitlab_codeclimate` and expects `1`status code
- launch binary with invalid `Dockerfile` with `-f gitlab_codeclimate` and `--no-fail` and expects `0`status code
- launch binary with valid `Dockerfile` with `-f gitlab_codeclimate` and expects `0`status code
- launch binary with valid `Dockerfile` with `-f gitlab_codeclimate` and `--no-fail` and expects `0`status code

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -137,7 +137,8 @@ exitProgram cmd res
 runLint :: CommandOptions -> Hadolint.LintOptions -> NonEmpty.NonEmpty String -> IO()
 runLint cmd conf files = do
   res <-  Hadolint.lint conf files
-  Hadolint.printResults (format cmd) res >>  exitProgram cmd res
+  Hadolint.printResults (format cmd) res
+  exitProgram cmd res
 
 main :: IO ()
 main = do
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -5,9 +5,10 @@
 module Main where
 
 import Control.Applicative
-import qualified Data.List.NonEmpty as NonEmpty
 import Data.Semigroup ((<>))
 import qualified Data.Set as Set
+import qualified Data.Sequence as Seq
+import qualified Data.List.NonEmpty as NonEmpty
 import Data.String (IsString (fromString))
 import qualified Data.Version
 import qualified Development.GitRev
@@ -41,6 +42,7 @@ import System.Exit (exitFailure, exitSuccess)
 
 data CommandOptions = CommandOptions
   { showVersion :: Bool,
+    noFail :: Bool,
     configFile :: Maybe FilePath,
     format :: Hadolint.OutputFormat,
     dockerfiles :: [String],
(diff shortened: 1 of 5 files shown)


app/Main.hs
```diff
@@ -6,8 +6,10 @@ module Main where
 
 import Control.Applicative
 import qualified Data.List.NonEmpty as NonEmpty
-import Data.Semigroup ((<>))
+import Data.Semigroup()
```
lorenzo, line 9: Please leave the line as it was, it is meant for older versions of haskell

app/Main.hs
```diff
@@ -6,8 +6,10 @@ module Main where
...
 import qualified Data.Set as Set
+import qualified Data.Sequence as Seq
+import Data.List.NonEmpty()
```
lorenzo, line 12: please use `qualified` and remove the parenthesis

app/Main.hs
```diff
@@ -68,13 +71,16 @@ parseOptions :: Parser CommandOptions
...
     <$> version -- CLI options parser definition
+    <*> noFail
     <*> configFile
     <*> outputFormat
     <*> files
     <*> lintOptions
   where
     version = switch (long "version" <> short 'v' <> help "Show version")
 
+    noFail = switch (long "no-fail" <> help "Do not fail when any rule is violated")
```
lorenzo, line 82: `Don't exit with a failure status code when any rule is violated`

src/Hadolint/Lint.hs
```diff
@@ -38,21 +37,15 @@ data OutputFormat
...
-        Codacy -> Codacy.printResult res >> exitSuccess
+printResults :: OutputFormat -> Format.Result Text DockerfileError -> IO ()
+printResults format allResults = do
+  case format of
+    TTY -> TTY.printResult allResults
+    Json -> Json.printResult allResults
+    Checkstyle -> Checkstyle.printResult allResults
+    CodeclimateJson -> Codeclimate.printResult allResults
+    GitlabCodeclimateJson -> Codeclimate.printGitlabResult allResults
+    Codacy -> Codacy.printResult allResults
```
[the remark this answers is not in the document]
antechrestos: @lorenzo can I knwo why? The code does not brings any info on it.

antechrestos:
@lorenzo is the implementation ok? Shall I need to change anything? Please let me know

lorenzo:
Thanks, it's great now :)

antechrestos:
Thanks @lorenzo

It's Sunday you should rest ☺

PR #530, opened 2021-02-02
Adding rule for maintainer label.

luckoseabraham:
### What I did
Added new rule for ensuring the presence of maintainer label.

The maintainer directive in docker was deprecated. But the feature itself was moved to using label directive and there are no rules for ensuring the presence of maintainer label.
This has been mentioned as best practice in couple of articles like https://snyk.io/blog/10-docker-image-security-best-practices/

For an organization, that has hundreds of team creating images, this would be a way to identify owners for remediation and governance.

### How I did it 
Added a new rule.

### How to verify it
Added test cases in spec.hs and tested using binary on dockerfile.

### DIFF
--- a/src/Hadolint/Rules.hs
+++ b/src/Hadolint/Rules.hs
@@ -12,6 +12,7 @@ import Data.List.NonEmpty (toList)
 import qualified Data.Map as Map
 import qualified Data.Set as Set
 import qualified Data.Text as Text
+import Data.Maybe (isJust, fromJust)
 import Data.Void (Void)
 import GHC.Generics (Generic)
 import qualified Hadolint.Shell as Shell
@@ -236,7 +237,8 @@ rules =
     dnfVersionPinned,
     pipNoCacheDir,
     noIllegalInstructionInOnbuild,
-    noSelfreferencingEnv
+    noSelfreferencingEnv,
+    hasMaintainerLabel
   ]
 
 optionalRules :: RulesConfig -> [Rule]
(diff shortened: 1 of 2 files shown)


src/Hadolint/Rules.hs
```diff
@@ -310,6 +312,41 @@ hasNoMaintainer = instructionRule code severity message check
     check (Maintainer _) = False
     check _ = True
 
+
+hasMaintainerLabel :: Rule
+hasMaintainerLabel dockerfile = instructionRuleState code severity message check [] dockerfile
+  where
+    code = "DL3012"
```
luckoseabraham, line 319: Have reused the DL3012 here since it makes sense from a description perspective. Not sure how to edit the wiki. 
If someone can let me know how to, I would be glad to take that on for providing an example.

lorenzo: the wiki is public for anyone to edit

lorenzo:
I'm not sure it is a great idea to suggest doing this in hadolint, do you have a use case? What is this preventing?

src/Hadolint/Rules.hs
luckoseabraham: Thanks @lorenzo  - will try editing it.

lorenzo:
@zemanlx what do you think of this idea?

lorenzo:
Thanks, @luckoseabraham, let us review the idea and figure out how to best introduce this change.

luckoseabraham:
> Thanks, @luckoseabraham, let us review the idea and figure out how to best introduce this change.

@lorenzo  - Thank you so much ! 👍

lorenzo:
There has been many requests in the past for checking labels and label structure. I wonder if we can put this check behind a flag that is meant for label checking.

what do you think?

luckoseabraham:
> There has been many requests in the past for checking labels and label structure. I wonder if we can put this check behind a flag that is meant for label checking.

I agree. 
Just to make sure that we are on the same page, the flag would be in the configuration yaml file and would look something like this ? 

```
ignored:
  - DL3000
  - SC1010

labelCheck: true
```

luckoseabraham:
> There has been many requests in the past for checking labels and label structure. I wonder if we can put this check behind a flag that is meant for label checking.
> 
> what do you think?

Hey @lorenzo  - As I was thinking more about the feature, the effort in https://github.com/hadolint/hadolint/pull/491 by @m-ildefons comes to mind. And the optional configuration makes more sense to me and could be a standard on all these niche rules that everyone is not forced up on.
I really dont want to duplicate the effort @m-ildefons has done on this aspect. And If that PR gets merged, I can easily switch to the severity Nothing.

What do you think ?

lorenzo:
That sounds like a good idea, I'll finish the remaining work there soon

luckoseabraham:
Hey @lorenzo  - I have made the rule optional using the severity DLIgnoreC. 
When you get a chance, Please take a look and let me know if you would like me to make any changes.

luckoseabraham:
Hey @lorenzo  - Following up to see if you have any inputs :)

src/Hadolint/Rules.hs
```diff
@@ -335,6 +337,41 @@ hasNoMaintainer = instructionRule code severity message check
     check (Maintainer _) = False
     check _ = True
 
+hasMaintainerLabel :: Rule
+hasMaintainerLabel dockerfile = instructionRuleState code severity message check [] dockerfile
+  where
+    code = "DL3012"
+    severity = DLIgnoreC
+    message = "Provide an email adress or URL as maintainer label"
```
lorenzo, line 345: `address` instead of `adress`

src/Hadolint/Rules.hs
```diff
@@ -335,6 +337,41 @@ hasNoMaintainer = instructionRule code severity message check
     check (Maintainer _) = False
     check _ = True
 
+hasMaintainerLabel :: Rule
+hasMaintainerLabel dockerfile = instructionRuleState code severity message check [] dockerfile
```
lorenzo, line 341: Can you please make the state a `Set Linenumber Text` instead of a plain list. Lists are not great for performance

lorenzo (commented):
Thanks, I made a few more comments

src/Hadolint/Rules.hs
luckoseabraham: Done. How silly of me :)

src/Hadolint/Rules.hs
luckoseabraham: Agree. 
I was looking for a consistent way of approaching a state of such scenarios. I see the map being used for another PR for health checks. 
So I have used the map here as well so that we can have a consistent approach for maintaining such state.

m-ildefons:
I really don't mean to thrash on your effort here, @luckoseabraham and the rule itself is great. But within the overall picture, this is going in the wrong direction IMO.
What I mean is that it will be too cumbersome and inflexible to represent label schemas as individual fixed rules. It is fine, if your whole schema is that there must be a label "maintainer", but it will not cut it when implementing schemata like the OCI annotations (https://github.com/opencontainers/image-spec/blob/master/annotations.md) or even label-schema.org labels.
Those would need dozens of rules each and Hadolint would be very slow to react, in case a schema is updated.
 Also imagine what the configuration would look like if you wanted to enable dozens of optional rules for label schema validation.
Do you see where I'm coming from?

My approach would be to build a rule generator, that reads the desired schema from the `.hadolint.yaml` and generates rules validating that schema. I've played around with some code in that direction, but it's not quite ready yet. I've made a draft [PR,](https://github.com/hadolint/hadolint/pull/563) so you can at least see what I'm on about and play around with it.

luckoseabraham:
hey @m-ildefons  - I have no issues with having a more flexible option be considered for label related logic.
The use case and implementation does make sense !

PR #533, opened 2021-02-05
bugfix: more accurate matching on `pip`

### LINKED ISSUE #511: DL3042 reports pipenv, however pipenv doesn't seem to support the option
- [x] This is a bug report

### Expected behavior

Only report `pip`.

### Actual behavior

`pipenv install` is reported by DL3042.

**Dockerfile**

```Dockerfile
FROM scratch

RUN pipenv install
```

Hadolint v1.19.0-45-gef91156-alpine

### LINKED ISSUE #532: Hadolint can't distinguish pip and pipx.
- [x] This is a bug report

pipx is a different tool from pip, it just starts with pip. It doesn't have the same cache switches and I don't think they do the same thing anyhow.

```
# hadolint ignore=DL3042
RUN	pipx install --no-cache "black==19.10b0"```

### PR DESCRIPTION
**Pull request by m-ildefons**
- Make matching pip commands more accurate

Pip has a lot of different ways of calling (`pip`, `pip3`, `python -m pip`,
...). Some rules are pip specific, while others apply to wrappers around
pip as well. These wrappers include `pipenv` and `pipx`. The addition of
a isPipWrapper command lets a rule distinguish between pip commands and
wrappers.

fixes #532
fixes #511

### What I did
I added an extensible way to distinguish between genuine `pip` commands and `pip` wrappers like `pipx`, when writing rules.
Simply excluding `pip` wrappers from matching should be up to the author of the individual rules, because a rule may apply to all `pip` commands including wrapper, or just the wrappers or just the genuine `pip` commands.

Use the `isPipWrapper` function to identify `pip` wrappers when linting through `pip` commands. It is extensible because it can easily be adapted to include the newest and most popular `pip` wrapper of the week.

### How to verify it
DL3042 no longer triggers with this Dockerfile:
```
RUN pipx install bla
RUN pipenv install bla
```

### DIFF
--- a/src/Hadolint/Rules.hs
+++ b/src/Hadolint/Rules.hs
@@ -1121,15 +1121,23 @@ pipNoCacheDir = instructionRule code severity message check
     check (Run (RunArgs args _)) = argumentsRule (Shell.noCommands forgotNoCacheDir) args
     check _ = True
     forgotNoCacheDir cmd =
-      isPipInstall cmd && not(usesNoCacheDir cmd)
+      isPipInstall cmd && not(usesNoCacheDir cmd) && not (isPipWrapper cmd)
     usesNoCacheDir cmd   = "--no-cache-dir" `elem` Shell.getArgs cmd
 
 isPipInstall :: Shell.Command -> Bool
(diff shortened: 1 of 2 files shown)


lorenzo:
thanks, this is great!

PR #543, opened 2021-02-07
Feature: Configurable rule severity

m-ildefons:
- Make rule severity overridable with any other severity.
- Remove unnecessary imports

This feature is implemented with a custom rule severity type
`DLSeverity` which encapsulates the possible severities, including the
possibility that a rule is ignored by default.

### What I did
The is my second attempt at the long requested feature of customizable rule severeties.

It is an alternative implementation of the customizable rule severities, without using a `Maybe Severity`. Instead this is implemented with a custom `DLSeverity` type. This also is not intermingled with another different feature, which should make reviewing much easier. That said, it is missing an example of an opt-in rule.

The `DLSeverity` type is constructed analogous to the `Severity` type from `Shellcheck`, which was used before. This achieves an easy mapping of `Severity` types to `DLSeverity` types.

If this is a preferred implementation over the one included in https://github.com/hadolint/hadolint/pull/491, then I will abandon that PR and rebase the healthcheck rules onto this. I will also update this PR with updated documentation, if the code itself becomes a candidate for merging.

Just like in PR #491, the format of the config file looks like this:
```yaml
override:
  error:
    - DLXXXX
  warning:
    - DLYYYY
  info:
    - DLAAAA
  style:
    - DLBBBB

ignored:
  - DLZZZZ
```

The commandline syntax would be like so:
```
$ hadolint --ignore=DLXXXX --error=DLYYYY --style=DLZZZZ --info=DLAAAA --warning=DLBBBB Dockerfile
```

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -5,11 +5,10 @@
 module Main where
 
 import Control.Applicative
-import Data.Semigroup ((<>))
-import qualified Data.Set as Set
-import qualified Data.Sequence as Seq
-import qualified Data.List.NonEmpty as NonEmpty
 import Data.String (IsString (fromString))
+import qualified Data.List.NonEmpty as NonEmpty
+import qualified Data.Sequence as Seq
+import qualified Data.Set as Set
 import qualified Data.Version
 import qualified Development.GitRev
 import qualified Hadolint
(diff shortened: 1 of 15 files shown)


lorenzo:
@m-ildefons I finally have the time to review this. Can you please rebase and fix the conflicts. The code changes seem good to me at first glance

m-ildefons:
Thanks for taking the time to review this. I feel kind of bad about dumping such large PRs onto you and I hope its not too much.

lorenzo:
thanks for taking the time to do this, I really appreciate it @m-ildefons

Issue #267, opened 2018-09-12
Color code the output


kenden:
- [] This is a bug report
- [x] This is a feature request
- [x] I searched existing issues before opening this one

### Expected behavior

In order to make the errors more obvious, it would be nice to color them in red in the output.
Like shellcheck does:

![image](https://user-images.githubusercontent.com/480586/44208550-f25e5180-a160-11e8-99f6-c10ab4364d96.png)

Then as a user, it's quicker to see a problem.

### Actual behavior
All the results are in black and white.
Errors are not quick to spot.

### Steps to reproduce the behavior

```bash
$ echo 'FROM ubuntu:latest' > Dockerfile
$ hadolint Dockerfile
Dockerfile:1 DL3007 Using latest is prone to errors if the image will ever update. Pin the version explicitly to a release tag
```

### Version
```bash
$ hadolint --version
Haskell Dockerfile Linter 1.12.0-no-git

$ uname -a
Darwin rectacted-mac07.local 17.7.0 Darwin Kernel Version 17.7.0: Thu Jun 21 22:53:14 PDT 2018; root:xnu-4570.71.2~1/RELEASE_X86_64 x86_64 i386 MacBookPro11,3 Darwin
```

kenden:
Maybe the colors could be used like they are in shellcheck:
https://github.com/koalaman/shellcheck/blob/f91b5bc2705b48d6436e92076dcf77866edcb518/src/ShellCheck/Formatter/TTY.hs#L38

In the example below, 
all the lines could be red.
or just the part about the error code (ex DL3009)?

When there is a lot of output, it would make it easier to find the errors.

```bash
$hadolint foldername/Dockerfile
foldername/Dockerfile:6 DL3009 Delete the apt-get lists after installing something
foldername/Dockerfile:6 DL3015 Avoid additional packages by specifying `--no-install-recommends`
foldername/Dockerfile:6 DL3014 Use the `-y` switch to avoid manual input `apt-get -y install <package>`
foldername/Dockerfile:98 DL3003 Use WORKDIR to switch to a directory
foldername/Dockerfile:100 DL3003 Use WORKDIR to switch to a directory
foldername/Dockerfile:107 DL3003 Use WORKDIR to switch to a directory
foldername/Dockerfile:109 SC2185 Some finds don't have a default path. Specify '.' explicitly.
foldername/Dockerfile:109 DL3003 Use WORKDIR to switch to a directory
[...]

PR #547, opened 2021-02-17
feature: colorize TTY output  #267

m-ildefons:
- add `--color` flag to switch on colorization
- default off
- colorize TTY output:
    error - bold red
    warning - bold yellow
    info - green
    style - cyan/blue

As elaborated in issue #267, this enhances the experience with Hadolint
on the console.
hadolint-colors

Since currently there are no rules with the `style` severity, I can not showcase its color.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -43,6 +43,7 @@ import System.Exit (exitFailure, exitSuccess)
 data CommandOptions = CommandOptions
   { showVersion :: Bool,
     noFail :: Bool,
+    color :: Bool,
     configFile :: Maybe FilePath,
     format :: Hadolint.OutputFormat,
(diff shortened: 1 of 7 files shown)


lorenzo:
This is great!!

A few things:
* Can you make it so that colorized in the default? That would mean flipping the flag to `--no-color`
* Honour the env variable `NO_COLOR` (https://no-color.org)
* Use https://github.com/kowainik/colourista so that you get windows compatibility for free

m-ildefons:
Thanks for the quick review. Colourista is new to me, but it does exactly, what I wanted and could not get from `System.Console.ANSI` directly.
I've also implemented your other two requests.

lorenzo:
Great job!

PR #548, opened 2021-02-17
rules: Add rule checking agains selfreferencing `ENV`s

### LINKED ISSUE #509: Add rule to check usage of variables defined in the same ENV statement

### Expected behavior

A warning is raised when a variable that is defined in the same ENV statement

### Actual behavior

No rule is triggered

**Dockerfile** 

```Dockerfile

ENV PACKAGE_NAME=${PACKAGE_NAME_ARG} \
  PACKAGE_BIN_DIR=/home/${PACKAGE_NAME}/vendor/bin

```

m-ildefons:
- Add Rule to check agains `ENV` statements where one variable
  references another variable defined within the same `ENV` statement.
- fixes #509
- Add tests for that rule

Docker will not expand environment variables within the same `ENV`
statement where they are defined.

The following Dockerfile will produce error DL3044:
```
ENV BLA="blubb" \
    BLUBB="${BLA}/blubb"
```
That is because Docker will expand the environment variable `BLUBB` to `"/blubb"`, not as expected to `"blubb/blubb"`. This behaviour could cause bugs and build errors in docker images and therefore Hadolint should notice these situations.

### DIFF
--- a/src/Hadolint/Rules.hs
+++ b/src/Hadolint/Rules.hs
@@ -223,7 +223,8 @@ rules =
     dnfCleanup,
     dnfVersionPinned,
     pipNoCacheDir,
-    noIllegalInstructionInOnbuild
+    noIllegalInstructionInOnbuild,
+    noSelfreferencingEnv
   ]
 
 optionalRules :: RulesConfig -> [Rule]
@@ -1164,3 +1165,21 @@ noIllegalInstructionInOnbuild = instructionRule code severity message check
     check (OnBuild (From _)) = False
     check (OnBuild (Maintainer _)) = False
     check _ = True
+
+
+noSelfreferencingEnv :: Rule
+noSelfreferencingEnv = instructionRule code severity message check
+  where
+    code = "DL3044"
+    severity = ErrorC
+    message = "Do not refer to an environment variable within the same `ENV` statement where it is defined."
+    check (Env pairs) = findReferences pairs
+    check _ = True
+
+    findReferences :: Pairs -> Bool
+    findReferences prs = null [ var | var <- map fst prs,
+                                      var `isSubstringOfAny` map snd prs ]
+    isSubstringOfAny :: Text.Text -> [Text.Text] -> Bool
+    isSubstringOfAny t l = not $ null [ v | v <- l,
+        (Text.pack "${" <> t <> Text.pack "}") `Text.isInfixOf` v
+        || (Text.pack "$" <> t) `Text.isInfixOf` v]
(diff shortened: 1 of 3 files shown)


src/Hadolint/Rules.hs
```diff
@@ -1164,3 +1165,19 @@ noIllegalInstructionInOnbuild = instructionRule code severity message check
...
+    severity = ErrorC
+    message = "Do not refer to an environment variable within the same `ENV` statement where it is defined."
+    check (Env pairs) = findReferences pairs
+    check _ = True
+
+    findReferences :: Pairs -> Bool
+    findReferences prs = null [ var | var <- map fst prs,
+                                      var `isSubstringOfAny` map snd prs ]
+    isSubstringOfAny :: Text.Text -> [Text.Text] -> Bool
+    isSubstringOfAny t l = not $ null [ v | v <- l, t `Text.isInfixOf` v]
```
lorenzo, line 1183: Would this trigger on partial matches? For example, I define `FOO` and then do `BAR = ${FOOBAR}`. Would this case trigger the rule incorrectly?

m-ildefons: Thats a good point. I was thinking about partial matches of the kind `BAR=/somepath/${FOO}`, but forgot your case. Anyways, I fixed it and added tests to cover this case as well.

lorenzo:
great job!

peterdeme:
@lorenzo this doesn't work with `ARGS`.

```dockerfile
ARG SERVICE_NAME

ENV SERVICE_NAME=${SERVICE_NAME}
```

chadlwilson:
Isn't this legitimate use if the variable is defined in a base image and you are attempt to redeclare it; e.g to prefix the `PATH` with something additional you are layering on?

m-ildefons:
@chadlwilson good point. It's beyond the scope of Hadolint atm. to introspect the base image. Therefore afaik it is impossible to know, what is and what isn't defined. Granted, `PATH` will be an exception, since it will always be defined.

The point of this rule is the case in the linked issue, where unexpectedly an environment variable is expanded to an empty string. This is especially unexpected, with two different variables.

I guess it might be an option to exempt cases where an environment variable is redefined with itself, since in that case one could assume the developer takes responsibility for making sure it gets expanded to the correct value.

What do you think, would this be a valid solution?

krispharper:
This is breaking our CI for exactly the above reason. We have lots of Dockerfiles that extend `PATH` with `ENV PATH /a/path:$PATH`. What's the recommended way to do that with `DL3044` in place?

chadlwilson:
@m-ildefons it sounds like a reasonable compromise to me to exempt cases where an env var refers to itself. Worst case scenario it's a typo and the user intended to use a different variable, but I don't think we can catch everything without false positives like this kind of case?

@krispharper `# hadolint ignore=DL3044` on preceding line, ftw (for now!) :-)

PR #550, opened 2021-02-20
Add rule for `COPY` when no `WORKDIR` is set  #390

### LINKED ISSUE #390: Find usage of relative COPY without WORKDIR
- [ ] This is a bug report

Today I fixed a bug caused by removing a WORKDIR instruction and not updating the paths of several COPY ones.

Now my suggestion would be to lint if a COPY destination is relative and no WORKDIR was set previously and let the user know that he should either set a WORKDIR or set a absolute path.

m-ildefons:
- Add rule finding instance of `COPY` with a relative destination when
  no `WORKDIR` has been set.
- Add tests for that rule

The rule permits not setting a `WORKDIR` in case the `WORKDIR` is
already set in a previous build stage (in the same Dockerfile)
form which it is inherited.
The rule always permits absolute destinations.

fixes  #390

### What I did
This Rule tracks whether or not a `WORKDIR` has been set in a build stage. In case a `COPY` with a relative destination occurs, but the `WORKDIR` has not been set (yet) in that build stage or any previous build stage that stage inherits from, it raises a warning.

The most complicated part is tracking which stage has a `WORKDIR` set, since they can inherit a `WORKDIR` from a previous stage and can have multiple `WOKRDIR` instructions. Therefore, the rule state accumulates a list of `(image or alias name, workdir)` pairs. In addition to that, it keeps track of which stage is is currently evaluated in.

### How to verify it
Tests are included.

This should raise a warning:
```Dockerfile
COPY foo bar
```

This should not raise a warning:
```Dockerfile
COPY ffoooo /bbaarr
WORKDIR /usr
COPY foo bar
```

Neither should this raise a warning
```Dockerfile
FROM debian:buster as base1
WORKDIR /usr
FROM base1 as base2
RUN foo
FROM base2
COPY ffoooo bbaarr
```

### DIFF
--- a/src/Hadolint/Rules.hs
+++ b/src/Hadolint/Rules.hs
@@ -11,6 +11,7 @@ import Data.List (foldl', isInfixOf, isPrefixOf, mapAccumL, nub)
 import Data.List.NonEmpty (toList)
 import Data.List.Index
 import qualified Data.Map as Map
+import Data.Maybe (isNothing, isJust, fromJust)
 import qualified Data.Set as Set
 import qualified Data.Text as Text
 import Data.Void (Void)
@@ -237,7 +238,8 @@ rules =
     dnfVersionPinned,
     pipNoCacheDir,
     noIllegalInstructionInOnbuild,
-    noSelfreferencingEnv
+    noSelfreferencingEnv,
+    relativeCopyWithoutWorkdir
   ]
 
 optionalRules :: RulesConfig -> [Rule]
@@ -1234,3 +1236,21 @@ noSelfreferencingEnv = instructionRuleState code severity message check Set.empt
         -- all characters valid in the inner of a shell variable name
         varChar :: String
         varChar = ['0'..'9'] ++ ['a'..'z'] ++ ['A'..'Z'] ++ ['_']
+
+relativeCopyWithoutWorkdir :: Rule
+relativeCopyWithoutWorkdir = instructionRuleState code severity message check ("", [])
+  where
+    code = "DL3045"
+    severity = DLWarningC
+    message = "`COPY` to a relative destination without `WORKDIR` set."
+    check st _ (From BaseImage {image, alias = Just als}) =
+        withState (unImageAlias als,
+                   snd st ++ [(unImageAlias als, dir) | (img, dir) <- snd st,
+                                                        img == imageName image]) True
+    check st _ (From BaseImage {image, alias = Nothing}) =
+        withState (imageName image, snd st) True
+    check st _ (Workdir dir) = withState (fst st, snd st ++ [(fst st, dir)]) True
+    check st _ (Copy (CopyArgs _ (TargetPath dest) _ _))
+      | fst st `elem` map fst (snd st) || "/" `Text.isPrefixOf` dest = withState st True
+      | otherwise = withState st False
+    check st _ _ = withState st True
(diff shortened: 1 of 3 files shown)


src/Hadolint/Rules.hs
```diff
@@ -1164,3 +1166,21 @@ noIllegalInstructionInOnbuild = instructionRule code severity message check
     check (OnBuild (From _)) = False
     check (OnBuild (Maintainer _)) = False
     check _ = True
+
+relativeCopyWithoutWorkdir :: Rule
+relativeCopyWithoutWorkdir = instructionRuleState code severity message check ("", [])
```
lorenzo, line 1171: I think the state would be more cleanly represented, and likely more performant, using a `Map Text (Set Text)` it will allow you to lookup by image alias and check in a single operation that a dir already exists in the Set

src/Hadolint/Rules.hs
```diff
@@ -1164,3 +1166,21 @@ noIllegalInstructionInOnbuild = instructionRule code severity message check
...
+  where
+    code = "DL3044"
+    severity = WarningC
+    message = "`COPY` to a relative destination without `WORKDIR` set."
+    check st _ (From BaseImage {image, alias})
```
lorenzo, line 1176: Try using a bit more of pattern matching: `(From BaseImage {image, alias = Just theAlias}`

m-ildefons: I did, but I am not sure this is any better than before. But its not any worse, so why not.

src/Hadolint/Rules.hs
m-ildefons: I made it even simpler by using a `Map Text Bool`, since this rule really only cares about whether or not a `WORKDIR` is set and not what that is set to.

lorenzo:
do you mind resolving the conflicts first?

lorenzo:
more conflicts :( Do you mind fixing them?

test/Spec.hs
```diff
@@ -908,6 +908,97 @@ main =
...
+                ]
+        in do
+          ruleCatches relativeCopyWithoutWorkdir dockerFile
+          onBuildRuleCatches relativeCopyWithoutWorkdir dockerFile
+      it "ok: `COPY` to relative destination if `WORKDIR` has been set in base image" $
+        let dockerFile =
+              Text.unlines
+                [ "FROM debian:buster as base",
+                  "WORKDIR /usr",
+                  "FROM debian:buster as stage-inbetween",
```
lorenzo, line 954: This case seems buggy. In order to inherit from a previous stage you need here:

```dockerfile
FROM base as stage-inbetween
```

This example should trigger the rule

test/Spec.hs
```diff
@@ -908,6 +908,97 @@ main =
...
+                ]
+        in do
+          ruleCatchesNot relativeCopyWithoutWorkdir dockerFile
+          onBuildRuleCatchesNot relativeCopyWithoutWorkdir dockerFile
+      it "ok: `COPY` to relative destination if `WORKDIR` has been set in base image - windows" $
+        let dockerFile =
+              Text.unlines
+                [ "FROM microsoft/windowsservercore as base",
+                  "WORKDIR c:\\system32",
+                  "FROM microsoft/windowsservercore as stage-inbetween",
```
lorenzo, line 967: same here

lorenzo (changes requested):

test/Spec.hs
lorenzo: oh, my bad. I misread the test case

lorenzo (approved):

lorenzo:
great work!

PR #552, opened 2021-02-21
regression: Fix regression   #551

### LINKED ISSUE #551: Regression introduced few hours ago
- [x] This is a bug report

This bug was introduced in https://github.com/hadolint/hadolint/pull/548

### Expected behavior
This should work fine:

```dockerfile
ARG SERVICE_NAME

ENV SERVICE_NAME=${SERVICE_NAME}
```

### Actual behavior
Throws error:
```
/dev/stdin:17 DL3044 error: Do not refer to an environment variable within the same `ENV` statement where it is defined.
```

**Dockerfile** (if relevant)

```Dockerfile
FROM python:3.8-alpine

ENV SERVICE_NAME=$SERVICE_NAME

CMD ["/bin/bash"]
```

m-ildefons:
- Keep track of what `ENV`s and what `ARG`s have been defined in
  rule DL3044.

When a variable is already defined, it is ok to reference it, even when
redefining that variable in an `ENV` statement.

The behavior described in #551 is tested for in unit tests.

This fix works by maintaining a list of defined variable names and excluding those when searching for occurrences of variables within the right hand side of the definition of variables. This list is fed by variables defined both in `ENV` and in `ARG` statements.

@lorenzo I think I have all possible cases of partial and full matches covered with the unit test, but please take a second and see if you can find some that I haven't covered yet.

fixes #551

### DIFF
--- a/src/Hadolint/Rules.hs
+++ b/src/Hadolint/Rules.hs
@@ -235,7 +235,8 @@ rules =
     dnfCleanup,
     dnfVersionPinned,
     pipNoCacheDir,
-    noIllegalInstructionInOnbuild
+    noIllegalInstructionInOnbuild,
+    noSelfreferencingEnv
   ]
 
 optionalRules :: RulesConfig -> [Rule]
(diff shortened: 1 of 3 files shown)


src/Hadolint/Rules.hs
```diff
@@ -1184,3 +1185,50 @@ noIllegalInstructionInOnbuild = instructionRule code severity message check
     check (OnBuild (From _)) = False
     check (OnBuild (Maintainer _)) = False
     check _ = True
+
+noSelfreferencingEnv :: Rule
+noSelfreferencingEnv = instructionRuleState code severity message check []
```
lorenzo, line 1190: Please use `Set` for the state. That way you avoid calling `nub`

m-ildefons: Done. The state is a Set now.

lorenzo:
let's give it a try

PR #554, opened 2021-02-24
rules: add rule linting `useradd`  #553

### LINKED ISSUE #553: Lint for `useradd` without `-l` flag
- [ ] This is a bug report

This is related to the follwing issue: https://github.com/moby/moby/issues/5419. As explaned there, docker can run into problems when using the command `useradd` without the `-l` option. This is because a large sparse logfile is created by that command and the overlay filesystem may have trouble dealing with that. The `-l` option disables the creation, thus avoiding the whole problem.

In my opinion, Hadolint should emit a warning in case the `-l` flag is not present, because
  * build times increase dramatically
  * image sizes increase dramatically

Given the age of the linked issue, it is unlikely, that this will be fixed anytime soon in upstream docker.

This behaviour is most noticeable when using large uids (>8 digits). This may be necessary in cases, where interaction with identity providers like FreeIPA is necessary.

I'll gladly look into writing a rule for this.

### Expected behavior
Hadolint should emit a warning.

### Actual behavior
No warning is emitted.

### Issue with Docker
To illustrate the necessity of having this warning, build an image with the following Dockerfile.
**Dockerfile** (if relevant)
```Dockerfile
FROM debian:buster
RUN useradd -u 123456789 runuser
```

```Shell
│ ~/tmp/breakage │ᐅ time docker build -t test .
Sending build context to Docker daemon  2.048kB
Step 1/2 : FROM debian:buster
 ---> ef05c61d5112
Step 2/2 : RUN useradd -u 123456789 runuser
 ---> Running in a73b006f4dc3
Removing intermediate container a73b006f4dc3
 ---> 7d5c4ae51acc
Successfully built 7d5c4ae51acc
Successfully tagged test:latest

real    10m36.483s
user    0m0.071s
sys     0m0.035s
│ ~/tmp/breakage │ᐅ docker images
REPOSITORY                                   TAG                        IMAGE ID       CREATED          SIZE
test                                         latest                     7d5c4ae51acc   2 minutes ago    40.1GB

```

`Haskell Dockerfile Linter v1.22.1-6-gbbfa5bb`

CentOS 7, Docker version 20.10.3, build 48d30b5

m-ildefons:
- Add rule to lint for missing flag `-l` during usage of `useradd` with
  long UIDs.
- Number is DL3045 in anticipation of DL3044 being merged
- Add tests for that rule

- Add function to exctract arguments to a flag from shell command arguments
- Add unit test module for the `Hadolint.Shell` module
- Add test the functions `getFlagArg` and `hasFlag`

fixes #553

# How I did it
This rule triggers when the argument following the `-u` flag of `useradd` is longer than 5 characters and the `-l` flag is absent.
The threshold of 5 was determined experimentally by increasing the UID until the resulting image size changed, but also matches the default UID_MAX value of 65535 for most Linux systems.

Also notable is the addition of unit tests for the `Hadolint.Shell` module.

### How to verify it
Either of these Dockerfiles should be ok:
```Dockerfile
RUN useradd -l -u 123456 luser
```
```Dockerfile
RUN useradd -u 12345 luser
```
This Dockerfile however should emit a warning:
```Dockerfile
RUN useradd -u 123456
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,6 +224,7 @@ Please [create an issue][] if you have an idea for a good rule.
 | [DL3043](https://github.com/hadolint/hadolint/wiki/DL3043)   | `ONBUILD`, `FROM` or `MAINTAINER` triggered from within `ONBUILD` instruction.                                                                      |
 | [DL3044](https://github.com/hadolint/hadolint/wiki/DL3044)   | Do not refer to an environment variable within the same `ENV` statement where it is defined.                                                        |
 | [DL3045](https://github.com/hadolint/hadolint/wiki/DL3045)   | `COPY` to a relative destination without `WORKDIR` set.                                                                                             |
+| [DL3046](https://github.com/hadolint/hadolint/wiki/DL3046)   | `useradd` without flag `-l` and high UID will result in excessively large Image.                                                                    |
 | [DL4000](https://github.com/hadolint/hadolint/wiki/DL4000)   | MAINTAINER is deprecated.                                                                                                                           |
 | [DL4001](https://github.com/hadolint/hadolint/wiki/DL4001)   | Either use Wget or Curl but not both.                                                                                                               |
 | [DL4003](https://github.com/hadolint/hadolint/wiki/DL4003)   | Multiple `CMD` instructions found.                                                                                                                  |
--- a/hadolint.cabal
+++ b/hadolint.cabal
@@ -4,7 +4,7 @@ cabal-version: 2.0
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: 06df88cbceaf76da3599a8943568f912b17a63d1795839eb4da974832e1ae0d4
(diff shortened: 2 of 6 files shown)


lorenzo:
do you mind resolving the conflicts first?

lorenzo:
sorry, more conflicts after I merged your other pull request. If you continue submitting more rules maybe we will need to split the rules in different files 😄

m-ildefons:
Thanks for reviewing this. I should really learn to make PRs that don't conflict with each other :smiley:

On the note of splitting the rules in different files, I don't think its worth the effort (yet). The only advantage I can see is that it would allow us to restructure the tests such that the tests are in the same file as the rules themselves without cluttering the source too badly.

lorenzo:
@m-ildefons can you please create the wiki page?

Thanks for your constant great work!

PR #555, opened 2021-02-25
rules: add `HEALTHCHECK` rules

### LINKED ISSUE #360: Check healthcheck presence

m-ildefons:
- Add DL4007 ensuring the presence of a `HEALTHCHECK` command.
  This rule is disabled by default
- Add DL4008 ensuring a build stage has one `HEALTHCHECK` command at
  most. This rule is enabled by default with severity `DLErrorC`

- Add tests for both rules

fixes #360

### What I did
As promised here https://github.com/hadolint/hadolint/pull/491#issuecomment-782609292, this is the rebase of the two healthcheck rules onto master. Little has changed, except that the unit tests are now separated by rule instead of intermixed.
The optional rule DL4007 has been updated to keep its state in a `Map Text.Text Bool` instead of a list of pairs now. I've also increased the use of pattern matching to separate the different conditions of this rule better.

You can enable the rule DL4007 by manually overriding the severity either on the command line:
`hadolint --warning=DL4007 Dockerfile` or in the config file:

```yaml
override:
  warning:
    - DL4007
```


src/Hadolint/Rules.hs
```diff
@@ -1232,3 +1234,48 @@ noSelfreferencingEnv = instructionRuleState code severity message check Set.empt
...
+      | otherwise = withState st False
+    check st line (From BaseImage {image, alias = Nothing})
+      | Map.findWithDefault False (imageName image) st = withState st True
+      | null (allHealthchecksInStage line dockerfile)
+          && null (allFromsAfter line dockerfile) = withState st False
+      | null (allHealthchecksInStage line dockerfile) = withState st True
+      | otherwise = withState st True
+    check st _ _ = withState st True
+
+    allHealthchecks df = [(l, h) | (l, Healthcheck h) <- instr df]
```
lorenzo, line 1260: This is perhaps more of a challenge than a request. 

In many other rules we use this pattern in which we traverse the same docker file multiple times to get the lines for certain instructions. But the line number you can get, anyways, from each `check` call.

Line numbers can be stored in the state, so in, theory, you only need to compare the current line in `check` to other lines in the state. Do you think that would be possible for you to do here?

m-ildefons: I'm all in for a challenge, but I'm not sure I quite grasp, what your on about yet.

I chose to use this pattern, because the case I want to detect here is the _absence_ of a `HEALTHCHECK`. In that case, I need a line to trigger the rule on, for which I chose the `FROM` line of the stage where the `HEALTHCHECK` is missing, because it is the only one guaranteed to be there and I found it easier to understand in cases of multi-stage builds, because it is close to the problem.
This means however, that I need a certain look-ahead to determine whether or not to trigger on that line, which I can't have, if I don't parse the Dockerfile.
The next thing is that determining if a `HEALTHCHECK` has been set in the stage(s) which produce(s) the final image(s) is more tricky than simply looking for a `HEALTHCHECK` command with a line number greater than a certain `FROM` line.

Anyways, I updated the patch set, because I found an edge case that was not correctly handled yet.

lorenzo: hmm, that makes sense, the current structure does not allow to set the error on any arbitrary line, just the line you are inspecting. I've been thinking of a refactor to change this fact, but that'll come later, I guess.

lorenzo:
changes look good, but there are conflicts again :(

m-ildefons:
I'll abandon this PR and make a new one with the changes on top of the master branch. Rebasing is not really worth it since almost everything in this PR needs to be changed.

PR #556, opened 2021-02-25
rules: refine selfreferencing env rule

m-ildefons:
- Allow a variable to reference itself in its definition once per `ENV`
  statement
- Add tests for new behavior

A variable may legitimately be redefined with itself e.g.
`PATH=/path:${PATH}`, thus this case should not trigger DL3044. However
doing so twice or more within the same `ENV` statement will not expand
the variable twice and therefore should be an error.

see https://github.com/hadolint/hadolint/pull/548#issuecomment-785800548

### How I did it
When searching for occurrences of the variable as reference (either $VAR or ${VAR}), the right hand side value of the variable itself is no longer searched. Only the right hand sides of other variables in the same `ENV` statement is searched. To implement this, an indexed list is used to uniquely identify the variable/value pairs. This can not be done by variable name, since a variable may be defined twice within the same `ENV` statement, which may result in wrong expansion of its value.

### How to verify it
The tests have been expanded to capture the new behavior.

This should no longer trigger DL3044:
```Dockerfile
ENV PATH=/bla:$PATH
```

This however should still trigger DL3044
```Dockerfile
ENV PATH=/bla:$PATH \
    PATH=/blubb:$PATH
```

### DIFF
--- a/hadolint.cabal
+++ b/hadolint.cabal
@@ -4,7 +4,7 @@ cabal-version: 2.0
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: 3ecbed4531d222b63b1238f288614505607a97e91af832440541cb8c7c678fa9
+-- hash: 06df88cbceaf76da3599a8943568f912b17a63d1795839eb4da974832e1ae0d4
 
 name:           hadolint
 version:        1.23.0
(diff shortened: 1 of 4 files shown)


lorenzo:
thanks!

PR #561, opened 2021-03-04
DL3045: fix when destination is a variable

### LINKED ISSUE #560: DL3045: false positive, missing documentation
- [x] This is a bug report

as a side note, didn't realize your Docker `latest` is linked to the Master branch versus the latest tagged release. 
Got broken pipelines because of https://github.com/hadolint/hadolint/commit/a556eafc42b10d188e098404e6ad617f90290f73

### Expected behavior

having locally:
```sh
> hadolint --version
Haskell Dockerfile Linter 1.23.0-no-git

> hadolint Dockerfile*

# i.e. all is fine
>
```

### Actual behavior

While our pipeline fails with the following:

```sh
<1s
+ hadolint --version
Haskell Dockerfile Linter v1.23.0-1-ga556eaf

hadolint Dockerfile*
<1s
+ hadolint Dockerfile*
Dockerfile:123 DL3045 warning: `COPY` to a relative destination without `WORKDIR` set.
Dockerfile:124 DL3045 warning: `COPY` to a relative destination without `WORKDIR` set.
Dockerfile:126 DL3045 warning: `COPY` to a relative destination without `WORKDIR` set.
Dockerfile:127 DL3045 warning: `COPY` to a relative destination without `WORKDIR` set.
Dockerfile:128 DL3045 warning: `COPY` to a relative destination without `WORKDIR` set.
Dockerfile:129 DL3045 warning: `COPY` to a relative destination without `WORKDIR` set.
...
```

image

Also, https://github.com/hadolint/hadolint/wiki/DL3045 doesn't exist and link is broken in README

**Dockerfile**

```Dockerfile
...

# Copy artefacts - lines 122-129
COPY --chown=${JETTY_USER}:${JETTY_USER} --from=jetty "${JETTY_BASE}" "${JETTY_BASE}"
COPY --chown=${JETTY_USER}:${JETTY_USER} --from=jetty /usr/local/jetty "${JETTY_HOME}"
COPY --chown=${JETTY_USER}:${JETTY_USER} --from=jetty docker-entrypoint.sh generate-jetty-start.sh /
COPY --chown=${JETTY_USER}:${JETTY_USER} --from=build /var/tmp/my_app/ "${JETTY_BASE}/webapps/${DEPLOYMENT_PREFIX}/"
COPY --chown=${JETTY_USER}:${JETTY_USER} my_app/ "${USER_HOME}/"
COPY --chown=${JETTY_USER}:${JETTY_USER} jetty/ "${JETTY_BASE}/"
COPY --chown=${JETTY_USER}:${JETTY_USER} webapps/ROOT/WEB-INF/ "${JETTY_BASE}/webapps/${DEPLOYMENT_PREFIX}/WEB-INF/"

# Container runtime settings
# reset back to Jetty user
USER ${JETTY_USER}
WORKDIR ${JETTY_BASE}
...
```

while the final `WORKDIR` is set after the `COPY` steps, but `hadolint` fails to recognize variables and whether they are using full or relative paths in them. In our case, all vars are with absolute paths.

m-ildefons:
- do not trigger DL3045 when the destination of the `COPY`
  statement is or begins with a variable.
- strip quotes from destination string

When the destination of a `COPY` statement is or begins with a
variable, it can be assumed, that the programer takes responsibility for
setting that variable.

fixes #560

### What I did
Exclude the case where the destination is a variable from DL3045. When looking at the original feature request https://github.com/hadolint/hadolint/issues/390, the case where the destination of the `COPY` statement is a variable is not really a problem this rule should warn about, since it is unaffected by unwanted interaction of the `WORKDIR` and `COPY` statements.
One can argue, that `COPY`ing to a destination that is a variable that has not been set or has been set to a relative directory both are situations to warn about, but that is perhaps not the scope of DL3045.

Also strip quotes from destination. This prevents false positives in cases like: `COPY src.sh "/destination"`

### How to verify it
Test cases included. This should no longer trigger DL3045:
```Dockerfile
FROM debian:buster
COPY src.sh ${DEST}
```
neither should this:
```Dockerfile
FROM debian:buster
COPY src.sh "/dest"
```

### DIFF
--- a/src/Hadolint/Rules.hs
+++ b/src/Hadolint/Rules.hs
@@ -1255,7 +1255,14 @@ relativeCopyWithoutWorkdir = instructionRuleState code severity message check ("
         withState (fst st, Map.insert (fst st) True (snd st)) True
     check st _ (Copy (CopyArgs _ (TargetPath dest) _ _))
         | uncurry Map.member st = withState st True  -- workdir has been set
-        | "/" `Text.isPrefixOf` dest = withState st True  -- absolute dest. normal
-        | ":\\" `Text.isPrefixOf` Text.drop 1 dest = withState st True  -- absolute dest. windows
+        | "/" `Text.isPrefixOf` Text.dropAround quotePredicate dest = withState st True  -- absolute dest. normal
+        | ":\\" `Text.isPrefixOf` Text.drop 1 (Text.dropAround quotePredicate dest) = withState st True  -- absolute dest. windows
+        | "$" `Text.isPrefixOf` Text.dropAround quotePredicate dest = withState st True -- dest is a variable
         | otherwise = withState st False
     check st _ _ = withState st True
+
+    quotePredicate :: Char -> Bool
+    quotePredicate c
+        | c == '"' = True
+        | c == '\'' = True
+        | otherwise = False
--- a/test/Spec.hs
+++ b/test/Spec.hs
@@ -912,18 +912,39 @@ main =
       it "ok: `COPY` with absolute destination and no `WORKDIR` set" $ do
         ruleCatchesNot relativeCopyWithoutWorkdir "COPY bla.sh /usr/local/bin/blubb.sh"
         onBuildRuleCatchesNot relativeCopyWithoutWorkdir "COPY bla.sh /usr/local/bin/blubb.sh"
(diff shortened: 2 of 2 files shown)


lorenzo:
looks good :)

PR #563, opened 2021-03-06
label-schema: Add configurable rules checking a label schema

### LINKED ISSUE #199: Ability to enforce values for LABEL via hadolint
[_This is a feature request, and I searched existing issues before opening this one._]

### Problem

In order to have better traceability of my container images, I'd like them to have various `LABEL`s on their `Dockerfile`s, e.g.:

- [`opencontainers`' `image-spec` labels](https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys), or
- [`label-schema` labels](https://microbadger.com/labels), or
- any other arbitrary label.

As human beings are notoriously unreliable and undeterministic, I'd like a way to to enforce such labels across projects' `Dockerfile`s.

### Potential solution

Since `hadolint` is a linter for `Dockerfile`s, maybe this should be:

- where such rules about labels can be defined, and
- how they should be enforced.

Thoughts?
N.B.: I currently do know Haskell, but fwiw, I'd be happy to try to come up with a PR.

### Alternatives

- [`dockerfile-lint`](https://github.com/projectatomic/dockerfile_lint/blob/master/sample_rules/recommended_label_rules.yaml)
- [`dli`](https://github.com/garethr/docker-label-inspector) (thanks for [pointing this out](https://github.com/hadolint/hadolint/issues/199#issuecomment-388076703) @zemanlx)

### LINKED ISSUE #480: Check for OCI compliant annotations/labels
- [] This is a bug report

### Expected behavior
Check to see if the Labels being used in your dockerfile follows the OCI stanard:
https://github.com/opencontainers/image-spec/blob/master/annotations.md

### Actual behavior
N/A

m-ildefons:
- Add configuration options to describe a label schema to Hadolint.
- Add commandline option --require-label, which can also describe a
  label schema to Hadolint
- Add a rule that alerts on any label that is not specified to be
  present
- Add a rule generator, which generates rules that check the existence
  of labels described in the label schema.
- Add a rule generator for generating rules that check that labels have
  a valid URL as value
- Add a rule generator for generating rules that check that labels have
  a valid time stamp according to RFC3339 as value
- Add a rule generator for generating rules that check that labels have
  a valid SPDX license identifier as value
- Add a rule generator for generating rules that check that labels have
  a valid git hash as value

fixes #480 
fixes #199 
conflicts with: #530 

This is a draft PR and not quite ready yet but please feel invited to play around with this and give me feedback.

Configure your label schema via config file like so:
```yaml
label-schema:
  my-label: text
  my-url-label: url
  my-license-label: spdx
  my-git-hash-label: hash
  my-rfc3339-compliant-date-label: rfc3339
```
or via commandline options:
```
$ hadolint --require-label my-label --requrie-label my-url-label:url --require-label my-license-label:spdx [...] Dockerfile
```
hadolint-labels

### Todo:
- [x] Make the superfluous label rule generator optional behind a --strict-labels flag, or similar
- [x] figure out tests
- [x] More informative messages
- [x] write documentation
- [x] Make the rule for label existence work better with multi stage builds
- [x] complete the value validation options, such that OCI labels can be fully validated
- [x] replace rule identifiers
- [x] Add rules for checking label names (see docker label inspector for example)

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE TemplateHaskell #-}
@@ -5,6 +6,7 @@
 module Main where
 
 import Control.Applicative
+import qualified Data.Bifunctor as Bifunctor
 import qualified Data.List.NonEmpty as NonEmpty
 import qualified Data.Map as Map
 import Data.Maybe
@@ -21,6 +23,7 @@ import Options.Applicative
     action,
     argument,
     completeWith,
+    eitherReader,
     execParser,
     fullDesc,
     header,
@@ -32,6 +35,7 @@ import Options.Applicative
     metavar,
     option,
     progDesc,
+    ReadM,
     short,
     showDefaultWith,
     str,
@@ -167,9 +171,9 @@ parseOptions =
         <*> ignoreList
         <*> parseRulesConfig
 
-    labels = Map.fromList . fmap labelParser
+    labels = Map.fromList
         <$> many
-          ( strOption
+          ( option readSingleLabelSchema
               ( long "require-label"
                   <> help "The option --require-label=label:format makes Hadolint check that the label `label` conforms to format requirement `format`"
                   <> metavar "LABELSCHEMA (e.g. maintainer:text)"
@@ -191,8 +195,16 @@ parseOptions =
               )
           )
 
-labelParser :: Text.Text -> (Rule.LabelName, Rule.LabelType)
-labelParser = Text.breakOn ":"
+type SingleLabelSchema = (Rule.LabelName, Rule.LabelType)
+
+readSingleLabelSchema :: ReadM SingleLabelSchema
+readSingleLabelSchema = eitherReader $ \s -> labelParser (Text.pack s)
+
+labelParser :: Text.Text -> Either String (Rule.LabelName, Rule.LabelType)
+labelParser l =
+    case Bifunctor.second (Rule.read . Text.drop 1) $ Text.breakOn ":" l of
+      (ln, Right lt) -> Right (ln, lt)
+      (_, Left e) -> Left $ Text.unpack e
 
 noFailure :: Hadolint.Result s e -> Bool
 noFailure (Hadolint.Result _ Seq.Empty Seq.Empty) = True
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -6,13 +6,16 @@ module Main where
 
 import Control.Applicative
 import qualified Data.List.NonEmpty as NonEmpty
+import qualified Data.Map as Map
 import Data.Maybe
 import qualified Data.Sequence as Seq
 import qualified Data.Set as Set
 import Data.String (IsString (fromString))
+import qualified Data.Text as Text
 import qualified Data.Version
 import qualified Development.GitRev
 import qualified Hadolint
+import qualified Hadolint.Rule as Rule
 import Options.Applicative
   ( Parser,
     action,
@@ -84,6 +87,9 @@ parseOptions =
     noFail = switch (long "no-fail" <> help "Don't exit with a failure status code when any rule is violated")
 
     nocolor = switch (long "no-color" <> help "Don't colorize output")
+
+    strictlabels = switch (long "strict-labels"
+        <> help "Do not permit labels other than specified in `label-schema`")
 
     configFile =
       optional
@@ -161,8 +167,22 @@ parseOptions =
         <*> ignoreList
         <*> parseRulesConfig
 
+    labels = Map.fromList . fmap labelParser
+        <$> many
+          ( strOption
+              ( long "require-label"
+                  <> help "The option --require-label=label:format makes Hadolint check that the label `label` conforms to format requirement `format`"
+                  <> metavar "LABELSCHEMA (e.g. maintainer:text)"
+              )
+          )
+
     parseRulesConfig =
-      Hadolint.RulesConfig . Set.fromList . fmap fromString
+      Hadolint.RulesConfig
+        <$> parseAllowedRegistries
+        <*> labels
+        <*> strictlabels
+
+    parseAllowedRegistries = Set.fromList . fmap fromString
         <$> many
           ( strOption
               ( long "trusted-registry"
@@ -170,6 +190,9 @@ parseOptions =
                   <> metavar "REGISTRY (e.g. docker.io)"
               )
           )
+
+labelParser :: Text.Text -> (Rule.LabelName, Rule.LabelType)
+labelParser = Text.breakOn ":"
 
 noFailure :: Hadolint.Result s e -> Bool
 noFailure (Hadolint.Result _ Seq.Empty Seq.Empty) = True
--- a/src/Hadolint/Config.hs
+++ b/src/Hadolint/Config.hs
@@ -2,19 +2,24 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE OverloadedStrings #-}
 
-module Hadolint.Config (applyConfig, ConfigFile (..), OverrideConfig (..)) where
+module Hadolint.Config
+  ( applyConfig,
+    ConfigFile (..),
+    OverrideConfig (..)
+  ) where
 
 import qualified Control.Foldl.Text as Text
 import Control.Monad (filterM)
 import qualified Data.ByteString as Bytes
 import Data.Coerce (coerce)
 import Data.Maybe (fromMaybe, listToMaybe)
+import Data.Map as Map
 import qualified Data.Set as Set
 import Data.YAML ((.:?))
 import qualified Data.YAML as Yaml
 import GHC.Generics (Generic)
 import qualified Hadolint.Lint as Lint
-import qualified Hadolint.Process
+import qualified Hadolint.Process as Process
 import qualified Hadolint.Rule
 import qualified Language.Docker as Docker
 import System.Directory
@@ -36,7 +41,9 @@ data OverrideConfig = OverrideConfig
 data ConfigFile = ConfigFile
   { overrideRules :: Maybe OverrideConfig,
     ignoredRules :: Maybe [Lint.IgnoreRule],
-    trustedRegistries :: Maybe [Lint.TrustedRegistry]
+    trustedRegistries :: Maybe [Lint.TrustedRegistry],
+    labelSchemaConfig :: Maybe Hadolint.Rule.LabelSchema,
+    strictLabelSchema :: Maybe Bool
   }
   deriving (Show, Eq, Generic)
 
@@ -54,6 +61,8 @@ instance Yaml.FromYAML ConfigFile where
     ignored <- m .:? "ignored"
     let ignoredRules = coerce (ignored :: Maybe [Text.Text])
     trustedRegistries <- m .:? "trustedRegistries"
+    labelSchemaConfig <- m .:? "label-schema"
+    strictLabelSchema <- m .:? "strict-labels"
     return ConfigFile {..}
 
 -- | If both the ignoreRules and rulesConfig properties of Lint options are empty
@@ -62,7 +71,7 @@ instance Yaml.FromYAML ConfigFile where
 -- return the error string.
 applyConfig :: Maybe FilePath -> Lint.LintOptions -> IO (Either String Lint.LintOptions)
 applyConfig maybeConfig o
-  | not (null (Lint.ignoreRules o)) && Lint.rulesConfig o /= mempty = return (Right o)
+  | not (Prelude.null (Lint.ignoreRules o)) && Lint.rulesConfig o /= mempty = return (Right o)
   | otherwise = do
     theConfig <-
       case maybeConfig of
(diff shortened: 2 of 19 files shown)


luckoseabraham:
have closed #530 in favor of this. So no more conflicts ! 👍

m-ildefons:
@lorenzo , I think its time to start the review on this PR.

There are a few behaviors, that differ slightly from the initial description up top:
- the rule `DL3047` checking for valid label keys is mandatory and independent of the label schema
- I've omitted requiring the reverse DNS notation for label keys, because labels with keys `maintainer` or `version` would not be conform, which would be overly strict IMO.
- if labels are present that are not in the label schema, no error is emitted by default, but `--strict-label` turns on reporting these cases
- you can make Hadolint enforce not having any labels present by omitting the label schema, but specifying `--strict-labels`
- There is an additional label type: `semver`, that works like `url` or `hash` above, but checks that the value of a label is a valid semantic version
- Hadolint will not do variable introspection, so if you wish to fill a label from an `ENV` or `ARG` variable, you need to set its value to `text` in the label schema or else it will be complained about.
- the generated rule include the label name that they triggered on in the message, where possible.

To check that a Dockerfile produces valid labels conforming to the OCI spec, this can be used in your `hadolint.yaml`:
```yaml
label-schema:
  org.opencontainers.image.created: rfc3339
  org.opencontainers.image.authors: text
  org.opencontainers.image.url: url
  org.opencontainers.image.documentation: url
  org.opencontainers.image.source: url
  org.opencontainers.image.version: semver
  org.opencontainers.image.revision: hash
  org.opencontainers.image.vendor: text
  org.opencontainers.image.licenses: spdx
  org.opencontainers.image.ref.name: text
  org.opencontainers.image.title: text
  org.opencontainers.image.description: text
```
This is the config for the equivalent but deprecated label-schema.org label schema:
```yaml
label-schema:
  org.label-schema.build-date: rfc3339
  org.label-schema.name: text
  org.label-schema.description: text
  org.label-schema.usage: url
  org.label-schema.url: url
  org.label-schema.vcs-url: url
  org.label-schema.vcs-ref: hash
  org.label-schema.vendor: text
  org.label-schema.version: semver
  org.label-schema.schema-version: semver
```

lorenzo:
@m-ildefons thanks for your great work. I’ll start the review soon. Please take a look at #565 which will create conflicts with this code, but I think you will be able to adapt what you have quickly.

src/Hadolint/Rules.hs
```diff
@@ -1284,3 +1292,178 @@ useraddFlagL = instructionRule code severity message check
...
+    check _ = True
+
+    hasNoInvalidKey prs = null $ [(l, v) | (l, v) <- prs,
+                                          Text.head l `notElem` ['a'..'z'] ||
+                                          Text.last l `notElem` ['a'..'z'] ++ ['0'..'9'] ||
+                                          containsIllegalChar l ||
+                                          hasReservedNamespace l ||
+                                          hasConsecutiveSeparators l]
+    containsIllegalChar :: Text.Text -> Bool
+    containsIllegalChar l = any (`notElem` validChars) (Text.unpack l)
```
lorenzo, line 1312: This could be ```Text.any (\c -> c `notElem` validChars)```

src/Hadolint/Rules.hs
```diff
@@ -1284,3 +1292,178 @@ useraddFlagL = instructionRule code severity message check
...
+    hasConsecutiveSeparators :: Text.Text -> Bool
+    hasConsecutiveSeparators l = ".." `Text.isInfixOf` l || "--" `Text.isInfixOf` l
+    validChars = ['0'..'9'] ++ ['a' .. 'z'] ++ ['.', '-']
+
+labelRuleGenerator :: Map.Map Text.Text Text.Text -> Bool -> [Rule]
+labelRuleGenerator labelschema strictlabels = concat
+  [
+    fmap missingLabelRule (Map.keys labelschema),
+    fmap labelIsNotEmptyRule (Map.keys labelschema),
+    fmap labelIsNotUrlRule (Map.keys (Map.filter (\v -> v == Text.pack "url") labelschema)),
```
lorenzo, line 1326: There is no need for `Text.pack` here, afaik. That's because of `OverloadedStrings`, which will convert automatically any string literal into the right type for you

src/Hadolint/Rules.hs
```diff
@@ -1284,3 +1292,178 @@ useraddFlagL = instructionRule code severity message check
...
+-- missingLabelRule
+--
+-- triggers on a `FROM` instruction when label `label` is not defined within
+-- that stage. Tracks defined labels through multi stage builds
+missingLabelRule :: Text.Text -> Rule
+missingLabelRule label dockerfile = instructionRuleState code severity message check ("", Map.empty) dockerfile
+  where
+    code = "DL3048"
+    severity = DLInfoC
+    message = Text.pack "Label `" <> label <> Text.pack "` is missing."
```
lorenzo, line 1343: same here, no need for `Text.pack`

src/Hadolint/Rules.hs
```diff
@@ -1284,3 +1292,178 @@ useraddFlagL = instructionRule code severity message check
...
+    getEmptyLabels :: (Eq l) => l -> [(l, Text.Text)] -> [(l, Text.Text)]
+    getEmptyLabels lbl prs = [(l, v) | (l, v) <- prs, l == lbl, Text.null v]
+
+labelIsNotUrlRule :: Text.Text -> Rule
+labelIsNotUrlRule label = instructionRule code severity message check
+  where
+    code = "DL3051"
+    severity = DLWarningC
+    message = Text.pack "Label `" <> label <> Text.pack "` is not a valid URL."
+    check (Label ls) = null $ getBadURLLabels label ls
```
lorenzo, line 1419: I magine this will trigger when having variables in the URL like `https://${DOMAIN}/my/thing`

src/Hadolint/Rules.hs
```diff
@@ -1284,3 +1292,178 @@ useraddFlagL = instructionRule code severity message check
...
+labelIsNotGitHashRule label = instructionRule code severity message check
+  where
+    code = "DL3054"
+    severity = DLWarningC
+    message = Text.pack "Label `" <> label <> Text.pack "` is not a valid git hash."
+    check (Label ls) = null $ getBadHashLabels label ls
+    check _ = True
+    getBadHashLabels lbl prs = [(l, v) | (l, v) <- prs, l == lbl, isBadHash v]
+    isBadHash :: Text.Text -> Bool
+    isBadHash h = any (`notElem` validHash) (Text.unpack h)
```
lorenzo, line 1456: could use `Text.any`

app/Main.hs
```diff
@@ -171,6 +187,11 @@ parseOptions =
               )
           )
 
+labelParser :: String -> (Text.Text, Text.Text)
+labelParser s = (head s', last s')
```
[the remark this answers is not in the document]
m-ildefons: That actually sounds a lot safer than what I am doing here.

src/Hadolint/Config.hs
```diff
@@ -80,15 +87,17 @@ applyConfig maybeConfig o
       contents <- Bytes.readFile configFile
       case Yaml.decode1Strict contents of
         Left (_, err) -> return $ Left (formatError err configFile)
-        Right (ConfigFile Nothing ignore trusted) ->
+        Right (ConfigFile Nothing ignore trusted labelschema strictlabels) ->
           return $
-            Right (applyOverride Nothing Nothing Nothing Nothing ignore trusted)
-        Right (ConfigFile (Just (OverrideConfig errors warnings infos styles)) ignore trusted) ->
+            Right (applyOverride Nothing Nothing Nothing Nothing ignore trusted labelschema strictlabels)
```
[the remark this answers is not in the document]
m-ildefons: Agreed. But in order to avoid blowing up this already large commit any further, I'd like not to do that within this PR.

src/Hadolint/Rules.hs
m-ildefons: Even easier than that, `Text.any (\`notElem\` validChars)` works just fine.

src/Hadolint/Rules.hs
m-ildefons: Thats right. This nicely reduces to `Map.filter (== "url")`

src/Hadolint/Rules.hs
m-ildefons: Yes indeed. That is the point. The literal string `${DOMAIN} is not valid as part of a URL. If you want to utilize variables, you would need to define that label as `text` in your schema.
It is a limitation of not being able to introspect variables.

src/Hadolint/Rule/DL3048.hs
```diff
@@ -0,0 +1,72 @@
...
+    markFail (StageID _ line) = CheckFailure {..}
+{-# INLINEABLE missingLabelRule #-}
+
+currentStage :: Text.Text -> StageID -> Acc -> Acc
+currentStage src stageid (Acc _ g b)
+    | not $ Set.null (Set.filter (predicate src) g) = Acc stageid (g |> Set.insert stageid) b
+    | otherwise = Acc stageid g (b |> Set.insert stageid)
+  where
+    predicate n0 StageID {name = n1} = n1 == n0
+currentStage _ stageid Empty = Acc stageid Set.empty (Set.empty |> Set.insert stageid)
```
lorenzo, line 68: `Set.empty |> Set.insert stageid` is just `Set.singleton stageid`

src/Hadolint/Rule/DL3047.hs
```diff
@@ -0,0 +1,35 @@
+{-# LANGUAGE OverloadedStrings #-}
```
lorenzo, line 1: You may remove all the language pragmas from `.hs` files as they are now globally enabled in `package.yaml`

lorenzo:
heads up, I just merged another rule with code `DL3047` so you need to shift numbers by one

src/Hadolint/Config.hs
```diff
@@ -123,13 +129,36 @@ applyConfig maybeConfig o
...
+      opts { Lint.rulesConfig =
+        ((`applyLabelSchema` labelschema) .
+        (`applyStrictLabels` strictlabels) .
+        (`applyTrustedRegistries` trusted)) (Lint.rulesConfig opts) }
+
+---    applyTrustedRegistries trusted opts
+---      | Prelude.null (Hadolint.Process.allowedRegistries (Lint.rulesConfig opts)) = opts
+---            --Hadolint.Process.allowedRegistries = toRules trusted
+---             -- <> (Hadolint.Process.allowedRegistries (Lint.rulesConfig opts))
+---      | otherwise = opts
```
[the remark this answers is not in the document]
m-ildefons: This was never meant to make it here. Oops :blush:

src/Hadolint/Process.hs
```diff
@@ -64,21 +74,27 @@ import qualified Hadolint.Shell as Shell
 import Language.Docker.Syntax
 
 -- | Contains the required parameters for optional rules
-newtype RulesConfig = RulesConfig
+data RulesConfig = RulesConfig
   { -- | The docker registries that are allowed in FROM
-    allowedRegistries :: Set.Set Registry
+    allowedRegistries :: Set.Set Registry,
+    labelSchema :: Map.Map Text.Text Text.Text,
```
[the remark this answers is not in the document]
m-ildefons: I see. I've decided to go the simple route for now and just make it a `type LabelType = Text.Text`. This avoids having to do some sort of parsing.
The parsing is not desirable IMO, because we would need to parse both the input from the command line options as well as from the config file. But I'll go the extra mile if you insist that a custom `data` type is worth it here.

src/Hadolint/Rule/DL3047.hs
```diff
@@ -0,0 +1,35 @@
...
+                                          containsIllegalChar l ||
+                                          hasReservedNamespace l ||
+                                          hasConsecutiveSeparators l]
+    containsIllegalChar :: Text.Text -> Bool
+    containsIllegalChar = Text.any (`notElem` validChars)
+    hasReservedNamespace :: Text.Text -> Bool
+    hasReservedNamespace l = "com.docker." `Text.isPrefixOf` l
+      || "io.docker." `Text.isPrefixOf` l
+      || "org.dockerproject." `Text.isPrefixOf` l
+    hasConsecutiveSeparators :: Text.Text -> Bool
```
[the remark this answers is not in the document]
m-ildefons: Done. Helpers are now top level functions with type signatures.

src/Hadolint/Rule/DL3047.hs
m-ildefons: This one is totally just a convenience thing for me, as my editor currently does not recognize global pragmas, but I think this may be a configuration issue on my end.

lorenzo: curious, what editor are you using?

lorenzo (commented):
looks great, just left a comment

src/Hadolint/Rule/DL3047.hs
m-ildefons: Vim with just a little bit of customization: I run a few auto commands, that remove trailing whitespaces from python sources and the likes, I also have `Ale` for running various linters and code checkers, and customized colors.
This particular problem was solvable by disabling everything in `Ale` that is not aware of either `package.yaml` or `*.cabal` files.

app/Main.hs
```diff
@@ -171,6 +191,9 @@ parseOptions =
               )
           )
 
+labelParser :: Text.Text -> (Rule.LabelName, Rule.LabelType)
```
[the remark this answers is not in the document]
m-ildefons: Current behavior is to silently ignore that and behave as if the user had specified `text` as the label type. I agree that this is not ideal.
 Let me try out how hard it is to implement parsing the label type. I'll try to combine it with making the label type a real custom data type instead of a alias to `Text.Text`.

src/Hadolint/Rule/DL3047.hs
lorenzo: Have you tried the haskell language server? I'm using it with (neo)vim and I'm extremely happy with it.

app/Main.hs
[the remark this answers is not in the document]
m-ildefons: This was finicky, but not too bad to implement.
Although yaml file do not support this type of custom data type very well, so the parser for the config file essentially abuses the yaml spec a bit, but the worst problem with that is that the yaml parser does not generate a fool proof error message, when parsing fails.
I've deliberately kept the behavior that not specifying a label type at all and specifying `text` is the same (i.e. no restriction on the content of the label), because this gives us a neat way to just require the presence of a label on the command line:
```
$ hadolint --require-label foo Dockerfile
```

lorenzo:
Great job!

I'm currently investigating why the compilation time for `Hadolint.Config` skyrocketed. We must have hit a bug or something.

lorenzo:
I'll merge first and try to find the culprit later :)

Thanks again for all your work!

PR #569, opened 2021-03-18
dl3044: fix bug with variable detection

### LINKED ISSUE #568: False positive DL3044
- [x] This is a bug report

### Expected behavior
```
ENV SAMTOOLS="$SOFT/samtools-$SAMTOOLS_VERSION/bin/samtools" \
    BGZIP="$SOFT/htslib-$SAMTOOLS_VERSION/bin/bgzip"
```
 should work as 
```
ENV SAMTOOLS="$SOFT/samtools-${SAMTOOLS_VERSION}/bin/samtools" \
    BGZIP="$SOFT/htslib-${SAMTOOLS_VERSION}/bin/bgzip"
```

### Actual behavior
```
ENV SAMTOOLS="$SOFT/samtools-$SAMTOOLS_VERSION/bin/samtools" \
    BGZIP="$SOFT/htslib-$SAMTOOLS_VERSION/bin/bgzip"
```
 throws `DL3044` while 
```
ENV SAMTOOLS="$SOFT/samtools-${SAMTOOLS_VERSION}/bin/samtools" \
    BGZIP="$SOFT/htslib-${SAMTOOLS_VERSION}/bin/bgzip"
```
not

m-ildefons:
When a variable is defined in terms of another variable and that other
variable is not referred to with curly-baces notation, the definition of
the first variable is split on the variable name. This results in a list
of remaining strings. The beginnings of the remaining strings is used to
determine, if the variable name that was found was just a substring of
another variable name, or is the actual variable. This is done by scanning
the beginnings of these strings for characters that terminate a variable name.
If one such character is found, the variable name looked for is actually found
and not just a substring of another variable name.
Since the `Text.splitOn` also returns the first part before the
variable, it is wrongfully determined, that the variable was found in
cases where its name was just a substring in another variable name.
Dropping this substring that never could begin with a terminator for a
variable name therefore fixes this problem.

fixes #568

This fixes #568 because the erroneous match is removed:
```Dockerfile
ENV SAMTOOLS="$SOFT/samtools-$SAMTOOLS_VERSION/bin/samtools" \
    BGZIP="$SOFT/htslib-$SAMTOOLS_VERSION/bin/bgzip"
           ^^^^^^^^^^^^^
           This part matches erroneously, therefore the rule thinks that the variable name is terminated after $SAMTOOLS and needs to be flagged.
```

### DIFF
--- a/src/Hadolint/Rule/DL3044.hs
+++ b/src/Hadolint/Rule/DL3044.hs
@@ -52,17 +52,17 @@ isSubstringOfAny t l =
 bareVariableInText :: Text.Text -> Text.Text -> Bool
 bareVariableInText v t =
   let var = "$" <> v
-      rest = Text.splitOn var t
+      rest = drop 1 $ Text.splitOn var t
    in var `Text.isInfixOf` t && any terminatesVarName rest
   where
     -- x would terminate a variable name if it was appended directly to
     -- that name
     terminatesVarName :: Text.Text -> Bool
-    terminatesVarName x = not $ beginsWithAnyOf x varChar
+    terminatesVarName x = Text.null x || not (beginsWithAnyOf x varChar)
 
     -- txt begins with any character of String
     beginsWithAnyOf :: Text.Text -> Set.Set Char -> Bool
-    beginsWithAnyOf txt str = Text.null txt || (Text.head txt `elem` str)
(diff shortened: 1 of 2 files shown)


lorenzo:
thanks!!

Issue #518, opened 2021-01-20
Unable to pull ghcr.io images


Iristyle:
- [x] This is a bug report
- [ ] This is a feature request
- [x] I searched existing issues before opening this one

### Expected behavior

Image pull from ghcr.io should work

### Actual behavior

Access denied - presumably due to images not being public?

### Steps to reproduce the behavior

```
docker pull ghcr.io/hadolint/hadolint:latest
Error response from daemon: Head https://ghcr.io/v2/hadolint/hadolint/manifests/latest: denied

❯ docker pull ghcr.io/hadolint/hadolint
Using default tag: latest
Error response from daemon: Head https://ghcr.io/v2/hadolint/hadolint/manifests/latest: denied

❯ docker pull ghcr.io/hadolint/hadolint:latest-alpine
Error response from daemon: Head https://ghcr.io/v2/hadolint/hadolint/manifests/latest-alpine: denied

❯ docker pull ghcr.io/hadolint/hadolint:latest-debian
Error response from daemon: Head https://ghcr.io/v2/hadolint/hadolint/manifests/latest-debian: denied
```

peaceiris:
We cannot find a correct docker image in release `1.21.0`, too.

https://github.com/hadolint/hadolint/runs/1801499660?check_suite_focus=true#step:5:83

lorenzo:
The publishing script is wrong, in a way I cannot figure out. Help appreciated

lorenzo:
This should be fixed for the upcoming v2.0.0, to be released in a few days

peaceiris:
Thank you for releasing v2.0.0!

The tag `2.0.0` is found, not `v2.0.0`. It seems to be different from the tag naming rule of v1.

```console
$ docker pull ghcr.io/hadolint/hadolint:v2.0.0
Error response from daemon: name unknown
$ docker pull ghcr.io/hadolint/hadolint:2.0.0 
2.0.0: Pulling from hadolint/hadolint
353680ae2880: Pull complete 
Digest: sha256:abd9d3803915b2992ba48a83bc462b2ddd0809689f2f80054c3b528ffb5eaa58
Status: Downloaded newer image for ghcr.io/hadolint/hadolint:2.0.0
ghcr.io/hadolint/hadolint:2.0.0
```

lorenzo:
hmm, that's not good. I'll look into that. Thanks!

lorenzo:
[...]

PR #504, opened 2020-11-30
Add workflow to deploy the container images to ghcr.io

angelbarrera92:
closes #503 

### What I did

I've created another workflow to build and deploy the container image to the GitHub Container Registry using the same scripts used by the DockerHub builder. This way, in case you need to modify the script, you don't have to worry about syncing the two builders.

### How I did it

I've created a secret under the repository named: `GHCR_IO_REGISTRY_TOKEN` with a personal access token with `write:packages` permissions. **Create it before triggering a new release.**

Captura de pantalla 2020-11-30 a las 16 32 17

Also, you will need to **enable the improved container support** under your `hadolint` organization settings:

Captura de pantalla 2020-11-30 a las 16 34 14

After the first release, you have to **mark the package as publicly accessible.**

Captura de pantalla 2020-11-30 a las 16 35 20

### How to verify it

I run the GitHub action in our SIGHUP's fork: https://github.com/sighupio/hadolint/runs/1474557615?check_suite_focus=true

PR #571, opened 2021-03-20
ci: improve Docker Release workflow

Juneezee:
### What I did

Setup the Docker Release workflow to build and push image to both Dockerhub and GitHub Container Registry on tag push.

This pull request aims to fix Issue #518.

This is an improvement to PR #504.

### How I did it

1. Renamed `.github/workflows/ghci.io.yml` into `.github/workflows/docker.yml`. This is because the Docker Release workflow now pushes to both **Dockerhub** and **GitHub Container Registry**. @lorenzo You will need to add `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN` as the repository secrets in order to push to Dockerhub.

2. Removed both `docker/hooks/build` and `docker/hooks/push`. The Docker Release workflow now uses the official [docker/build-push-action](https://github.com/docker/build-push-action) for building and pushing image. 

3. Updated [upx/upx](https://github.com/upx/upx) from `3.94` to `3.96`.

4. Correctly tag and push `scratch` image. Previously only `debian-distro` and `alpine-distro` was targeted and tagged. This is fixed by removing the `AS distro` from `FROM scratch AS distro`. Now the `scratch`, `alpine`, and `debian` images will be built and pushed correctly by the workflow. The image below shows the size of `scratch`, `alpine`, and `debian` image respectively.
image

5. For example, if the tag pushed is `v2.0.0`, 
    - `scratch` image will be tagged as `latest` and `2.0.0`
    - `alpine` image will be tagged as `latest-alpine` and `2.0.0-alpine`
    - `debian` image will be tagged as `latest-debian` and `2.0.0-debian`

### How to verify it

I have run the Docker Release workflow on my fork (https://github.com/Juneezee/hadolint/actions/runs/670702218).

image

#### Scratch image
image

#### Alpine image
image

#### Debian image
image


.github/workflows/docker.yml
```diff
@@ -0,0 +1,64 @@
...
+          platforms: linux/amd64
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
```
Juneezee, line 30: @lorenzo  You will need to add DOCKERHUB_USERNAME and DOCKERHUB_TOKEN as the repository secrets in order to push to Dockerhub.

Juneezee:
@lorenzo In addition to #518, I think this is likely because the package visibility has not been set to **public** yet.

This is what I see when I navigate to the `Packages` section of `hadolint` organisation. There is no `hadolint` package listed in there.

image

---

It should show the `hadolint` package if the package visbility is set to public. For example, the image below is one of the organisation that I own and has two public packages:

image

According to [Configuring visibility of container images for an organization](https://docs.github.com/en/packages/guides/configuring-access-control-and-visibility-for-container-images#configuring-visibility-of-container-images-for-an-organization), the default visibility is private unless it is explicitly set to public.

lorenzo:
@Juneezee I just made it public. Thanks for the guidance :)

lorenzo:
@Juneezee Currently, due to what I believe is a bug, we are pushing `latest` as whatever gets merged into master. Do you think we should keep that or just mark `latest` for actual tags?

Juneezee:
> @Juneezee Currently, due to what I believe is a bug, we are pushing `latest` as whatever gets merged into master. Do you think we should keep that or just mark `latest` for actual tags?

This is no longer the case after this pull request is merged. According to line 3 - 7 in `.github/workflows/docker.yml`,

```
name: Docker Release

on:
  # Trigger the workflow on the new 'v*' tag created
  push:
    tags:
      - "v*"
```

This workflow will only run when a tag starting with `v` has been pushed. I have configured the workflow in a way such that the `v` prefix will be remove when tagging the image. For example, if the tag `v2.0.0` is pushed, the workflow will tag the `scratch` image as `hadolint/hadolint:latest` and `hadolint/hadolint:2.0.0`.

lorenzo:
perfect, thanks so much for your help!

.github/workflows/docker.yml
```diff
@@ -0,0 +1,64 @@
...
+        run: |
+          if [ "${{ matrix.distro }}" = 'scratch' ]; then
+            echo ::set-output name=target::''
+            echo ::set-output name=latest::latest
+            echo ::set-output name=tag::"${GITHUB_REF/refs\/tags\/v/}"
+          else
+            echo ::set-output name=target::"${{ matrix.distro }}-distro"
+            echo ::set-output name=latest::"latest-${{ matrix.distro }}"
+            echo ::set-output name=tag::"${GITHUB_REF/refs\/tags\/v/}-${{ matrix.distro }}"
+          fi
```
Juneezee, line 50: This is where the workflow will set the target and tag for `docker build` command appropriately.

Issue #573, opened 2021-03-21
broken COPY in ONBUILD context


djalal:
- [x] This is a bug report
- [ ] This is a feature request
- [x] I searched existing issues before opening this one

### Expected behavior

hadolint should see "ONBUILD WORKDIR ..." has been set, and allow for a relative <dst> path in "ONBUILD COPY <src> <dst>"

### Actual behavior

check fails with 

```
-:7 DL3045 warning: `COPY` to a relative destination without `WORKDIR` set.
```

### Steps to reproduce the behavior

regression seems to be introduced first by version v1.23.0-4-g4db396c, version 1.23.0-3 works fine.

> docker run --rm -i hadolint/hadolint:v1.23.0-4-g4db396c

> cat Dockerfile
```
FROM python:3.9

ONBUILD WORKDIR /usr/src/app
ONBUILD COPY requirements.txt .
```

**Output of `hadolint --version` or
  `docker run --rm hadolint/hadolint hadolint --version` or
  `docker run --rm ghcr.io/hadolint/hadolint hadolint --version`:**

```bash
Haskell Dockerfile Linter v1.23.0-4-g4db396c
```

**Dockerfile** (if relevant)

```Dockerfile
FROM python:3.9

ONBUILD WORKDIR /usr/src/app
ONBUILD COPY requirements.txt .
```

**Additional environment details (OS, stack version, etc.)**

m-ildefons:
Thanks for the bug report, @djalal 
I've created a fix and it should be available in `master` in a few days.

PR #575, opened 2021-03-23
DL3045: Fix workdir detection in `ONBUILD` context

m-ildefons:
- Fix workdir detection on `ONBUILD` context by allowing the current
  stage to take the special value `None`, since an `ONBUILD` context has
  no current build stage.
- Add test cast specifically testing this case

fixes: #573

### What I did
With the larger restructuring, the way `ONBUILD` instructions are linted has changed. Now the rules are executed on a filtered version of the dockerfile, only containing the lines with `ONBUILD` instructions. This strips the `FROM` instructions out of the input to the rules, breaking rules that rely on finding at least one `FROM` instruction to initialize.
This fix concerns only rule DL3045 by modifying its state to allow for the possibility that the current context has no `FROM` instruction.

### How to verify it
The following dockerfile should no longer trigger DL3045
```Dockerfile
FROM python:3.9

ONBUILD WORKDIR /usr/src/app
ONBUILD COPY requirements.txt .
```

### DIFF
--- a/src/Hadolint/Rule/DL3045.hs
+++ b/src/Hadolint/Rule/DL3045.hs
@@ -8,13 +8,19 @@ import qualified Data.Text as Text
 import Hadolint.Rule
 import Language.Docker.Syntax
 
+-- This data encapsulates the name of a build stage. It may be None withing an
+-- `ONBUILD` context.
+data Stage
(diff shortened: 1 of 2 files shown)


lorenzo:
perfect, thanks!

PR #576, opened 2021-03-23
README.md: Add entries for DL3012 and DL3057

m-ildefons:
These entries somehow got lost or forgotten. This commit adds them in
their current correct form to the readme.


lorenzo:
thanks

PR #579, opened 2021-03-25
Adding rule DL3058 for email address format in label schema.

luckoseabraham:
### What I did
Created a new rule DL3058 for email format validation with label schema.
This is done to manage email values for maintainer label schema.

### How I did it
Added a new rule DL3058

### How to verify it
Added test cases in Spec.hs


luckoseabraham:
@lorenzo @m-ildefons - Will make an entry on the wiki corresponding to the link for rule in README once the PR is approved/merged.

Let me know of your thoughts.

Thanks !

src/Hadolint/Rule/DL3058.hs
```diff
@@ -0,0 +1,27 @@
...
+import Hadolint.Rule
+import Language.Docker.Syntax
+import Data.Text.Encoding
+
+
+rule :: LabelSchema -> Rule args
+rule labelschema = mconcat $
+  fmap labelIsNotEmailRule (Map.keys (Map.filter (== Email) labelschema))
+{-# INLINEABLE rule #-}
+{-# LANGUAGE OverloadedStrings #-}
```
m-ildefons, line 14: This should not be needed, as the `OverloadedStrings` compiler option is enabled globally.

src/Hadolint/Rule/DL3058.hs
```diff
@@ -0,0 +1,27 @@
...
+labelIsNotEmailRule label = simpleRule code severity message check
+  where
+    code = "DL3058"
+    severity = DLWarningC
+    message = "Label `" <> label <> "` is not a valid email format - must be conform to RFC5322."
+    check (Label ls) = null $ getBadEmailformatLabels label ls
+    check _ = True
+
+getBadEmailformatLabels :: LabelName -> Pairs -> Pairs
+getBadEmailformatLabels lbl pairs = [(l, v) | (l, v) <- pairs, l == lbl, not (Email.isValid (encodeUtf8 v))]
```
m-ildefons, line 27: Just a suggestion: `isValid` only accepts fully compliant strings. If you want to allow something like `user . name @ host . com` as well (with spaces), the `Text.Email.Validate` module has a `canonicalizeEmail` function, which can be used for that. I don't know if its still a common practice to obfuscate emails like that to avoid spam, but `canonicalizeEmail` would allow that too.

src/Hadolint/Rule/DL3058.hs
luckoseabraham: Thanks for noticing that. I added that for testing something and forgot to remove it. How silly of me ! 
Have removed it.

src/Hadolint/Rule/DL3058.hs
luckoseabraham: I dont have a use case for obfuscated emails and I personally dont like to make code change for such a rare use case that might or might not come.
how about we leave it be for now and if there is an ask, we can add it ? 
Or if you would like that added, I can definitely add that change too.

hadolint.cabal
```diff
@@ -3,8 +3,6 @@ cabal-version: 2.0
 -- This file has been generated from package.yaml by hpack version 0.34.2.
 --
 -- see: https://github.com/sol/hpack
---
--- hash: 37f589477154f969b62b2232bb0d1b63a8261cdc7c3b6099f0dcaad7240bc2e7
```
lorenzo: This line should not be deleted. Let stack generate this file for you automatically

luckoseabraham: Hey @lorenzo  - I did not delete the hash manually. This seems to be the default behavior for hpack 0.34.2 
https://hackage.haskell.org/package/hpack-0.34.0/changelog

For now, I will downgrade my stack version to use the 0.33 hpack so that the hash is still generated. But its bound to come our way.

src/Hadolint/Rule/DL3058.hs
m-ildefons: Fair enough, it was just a suggestion.

hadolint.cabal
lorenzo: I see, did not know a new version was out. What version where you using?

lorenzo:
Thanks so much for your contribution!

lorenzo:
@luckoseabraham can you please add the entry to the wiki?

luckoseabraham:
> @luckoseabraham can you please add the entry to the wiki?

Done :  https://github.com/hadolint/hadolint/wiki/DL3058

hadolint.cabal
luckoseabraham: I dont remember the exact stack version. But the hpack version was 0.34.2 which has the default of no hash strategy.

lorenzo:
Thanks!

PR #581, opened 2021-03-27
DL3045: Fix alternative windows paths #580

### LINKED ISSUE #580: DL3045: False positive on windows file path
- [x] This is a bug report

### Expected behavior

Using a hadolint docker image (hadolint/hadolint:v1.23.0-8-gb01c5a9-alpine)

When linting the following 
`COPY . C:/test`

No warnings/errors are given

### Actual behavior

Using a hadolint docker image (hadolint/hadolint:v1.23.0-8-gb01c5a9-alpine)

When linting the following 
`COPY . C:/test`

The following is 
./Dockerfile:9 DL3045 warning: `COPY` to a relative destination without `WORKDIR` set.

Hadolint doesn't seem to process windows paths using forward slashes (which though technically incorrect still work)
To reproduce run hadolint over this docker file
``` Dockerfile
FROM mcr.microsoft.com/windows/nanoserver:20H2

COPY . C:/test
COPY . C:\test
```
And see that it picks up one of the first line but not the second.

This is a vary minor issue and not a blocker (just change your slashes of course) but letting you know anyway

```bash
Haskell Dockerfile Linter v1.23.0-8-gb01c5a9
```

Windows 10 PC, running linux hadolint container (hadolint/hadolint:v1.23.0-8-gb01c5a9-alpine)

m-ildefons:
- Fix alternative windows paths
- Split off tests into its own file

fixes #580

The most impactful change here is, splitting off the unit tests for DL3045 into its own file, as this creates the infrastructure for doing so with the other tests as well. That makes it easier to write new tests, especially when creating new rules.

### How I did it
This just allows patterns like `c:/` under rule DL3045 as well.

### How to verify it
This should no longer trigger DL3045:
```Dockerfile
COPY . C:/foo
```


lorenzo:
Nice, I like the idea of splitting the tests

PR #582, opened 2021-03-27
Split test into dedicated files

m-ildefons:
### What I did
- Split up unit tests into dedicated files for each rule.
- Simplified pragma handling for tests

### How I did it
- move helper functions into their own module
- move each rule's tests into a new file just for that rule
- generalize compiler pragmas in `packages.yaml`

### How to verify it
Given how humongously large Spec.hs has grown, it also had become harder to handle and generally just a bear to work with. It also became increasingly hard to determine whether or not a particular rule was tested against a particular example, as sometimes these tests were strewn about in different sections of the file.

It is still ok for some tests to be directly in Spec.hs as they are concerned with the general functionality of the program, like testing inline ignores. Other tests - especially for individual rules - are moved to their individual files, giving a much easier access for developers.

All unit tests should still pass.


lorenzo:
great job!

Issue #494, opened 2020-11-18
code = "DL3000" false positive for windows paths


Pachoncilla:
- [x] This is a bug report
- [ ] This is a feature request
- [x] I searched existing issues before opening this one

### Expected behavior
No warning

### Actual behavior
Warning DL3000 Use absolute WORKDIR

### Steps to reproduce the behavior
docker run --rm -i hadolint/hadolint < Dockerfile

**Output of `hadolint --version` or
  `docker run --rm hadolint/hadolint hadolint --version`:**

Haskell Dockerfile Linter v1.18.2-19-gcd5fca8

```bash
/dev/stdin:5 DL3000 Use absolute WORKDIR
```

**Dockerfile** (if relevant)

```Dockerfile
FROM mcr.microsoft.com/dotnet/framework/sdk:4.8
WORKDIR 'c:\' 
```

**Additional environment details (OS, stack version, etc.)**

blackliner:
Relevant code: https://github.com/hadolint/hadolint/blob/master/src/Hadolint/Rules.hs#L291

I don't know haskell, but I think a regex would fit here? `^[\"\']?[a-zA-Z]:\\[\"\']?`
https://regex101.com/r/6vJ0nR/1

darkvertex:
Came here to report this then saw this issue.

Happened to me too with the slashes the other way sans quotes:
```Dockerfile
WORKDIR C:/myproj
```

m-ildefons:
Thanks for reporting this issue. While I was looking for a fix, I found another related issue in rule DL3045 as well. This should be fixed now.

PR #583, opened 2021-03-28
DL3000, DL3045:  fix absolute windows paths

m-ildefons:
### What I did
- Add recognition for absolute windows paths to DL3000
- Fix recognition for absolute windows paths for DL3045, which was
   broken because it did not recognize an absolute path at the top level
   like `C:\`
- Add tests for DL3000
    
fixes: #494


lorenzo:
needs rebasing now :)

PR #586, opened 2021-03-30
DL3016: fix mistakenly interpret argument to flag as package

### LINKED ISSUE #515: DL3016 "Pin versions in npm" misfires on flag args
- [x] This is a bug report

### Expected behavior

When I use `npm install --loglevel verbose` in a Dockerfile, hadolint should not complain about DL3016.

### Actual behavior

It does.

```bash
cat <<eof > Dockerfile
FROM ubuntu@20.04
RUN npm install --loglevel verbose
eof

hadolint Dockerfile
```

Has output:

```bash
Dockerfile:2 DL3016 Pin versions in npm. Instead of `npm install <package>` use `npm install <package>@<version>`
```

**Output of `hadolint --version`:**

```bash
$ hadolint --version
Haskell Dockerfile Linter 1.18.0-no-git
```

**Dockerfile**

```Dockerfile
FROM ubuntu@20.04
RUN npm install --loglevel verbose
```

macos (probably irrelevant)

m-ildefons:
- Add unit test case for `Shell.dropFlagArg`
- Add unit test case for DL3016 with `--loglevel verbose`
- make DL3016 ignore the arguments to named flags and not mistage them
  as packages
- move helper functions for DL3016 to top level and name their type
  signatures
- fixup `Shell.getFlagArg` to use a `Set` instead of a raw list
- remove superfluous use of `Text.pack` from `test/Shellspec.hs`

fixes #515

As I don't have very much experience with `npm`, there may very well be other flags that should have their arguments ignored as well, but from a quick glance over the documentation, I could not make out any.

### How to verify it

The following Dockerfile should no longer emit `DL3016`:
```Dockerfile
RUN npm --loglevel verbose install foo@1.0.0
```

### DIFF
--- a/src/Hadolint/Rule/DL3016.hs
+++ b/src/Hadolint/Rule/DL3016.hs
@@ -26,36 +26,63 @@ rule = simpleRule code severity message check
 
     check (Run (RunArgs args _)) = foldArguments (Shell.noCommands forgotToPinVersion) args
     check _ = True
-
-    forgotToPinVersion cmd =
-      isNpmInstall cmd && installIsFirst cmd && not (all versionFixed (packages cmd))
-
-    isNpmInstall = Shell.cmdHasArgs "npm" ["install"]
-    installIsFirst cmd = ["install"] `isPrefixOf` Shell.getArgsNoFlags cmd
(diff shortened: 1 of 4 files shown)


lorenzo:
thanks!

Issue #395, opened 2020-02-15
Suggest ``yarn cache clean`` when ``yarn [install]`` was run


SuperSandro2000:
- [ ] This is a bug report
- [x] This is a feature request
- [x] I searched existing issues before opening this one

### Suggestion

Yarn can generate quite large caches that can easily be 100MB or 300MB big.
I think hadolint should tell you to use ``yarn cache clean`` in the same layer you installed something to avoid keeping those caches in the final image.

ghostsquad:
Or use a multiple stages, which is likely better anyway, since you may need to install dev dependencies, then uninstall those before finalizing the final image.

PR #592, opened 2021-04-03
DL3060: Lint for `yarn cache clean` #395

m-ildefons:
- Add new rule DL3060 linting for `yarn cache clean` after `yarn
  install` with severity `DLInfoC`
- Add tests

This rule reminds the odd programmer about cleaning `yarn`s caches after
installing, just like with the other package managers.

fixes: #395
### How to verify it
This should trigger `DL3060`:
```Dockerfile
RUN yarn install foo
```
This should be fine:
```Dockerfile
RUN yarn install foo \
 && yarn cache clean
```


lorenzo:
this looks great, thanks!

Issue #593, opened 2021-04-05
json "message" field incorrect for "missing whitespace" error


infokiller:
- [x] This is a bug report
- [x] I searched existing issues before opening this one

### Expected behavior

Clean error message in "message" field: `missing whitespace`

### Actual behavior

Error message mixed with other data: `/dev/stdin:1:4:\n  |\n1 | RUNNN\n  |    ^\nmissing whitespace\n`

### Steps to reproduce the behavior

```bash
docker run --rm -i hadolint/hadolint hadolint -f json - < <(echo 'RUNNN')
```

**Output of `hadolint --version` or
  `docker run --rm hadolint/hadolint hadolint --version` or
  `docker run --rm ghcr.io/hadolint/hadolint hadolint --version`:**

```bash
Haskell Dockerfile Linter v2.1.0-0-g85a5357
```

**Dockerfile** (if relevant)

```Dockerfile
RUNNN
```

**Additional environment details (OS, stack version, etc.)**

Linux 5.11.11-hardened1-1-hardened

PR #594, opened 2021-04-05
json-formatter: fix error message

m-ildefons:
Fix `message` field in output by json formatter in case of errors
preventing rule application (e.g. parsing errors).

Previously, a pretty printing method from `Text.Megaparsec.Error` was
used, but this decorated the message with junk characters. The new
`errorMessage` function converts a `ParseErrorBundle` into a printable
`String` without additional junk, stripping newlines from the end as
well.

fixes: #593

### How to verify it
Executing `hadolint -f json Dockerfile` for the following Dockerfile should yield `[{"line":1,"code":"DL1000","message":"missing whitespace","column":4,"file":"/home/moritz/tmp/Dockerfile","level":"error"}]`:
```Dockerfile
RUNNN
```

### DIFF
--- a/src/Hadolint/Formatter/Format.hs
+++ b/src/Hadolint/Formatter/Format.hs
@@ -1,6 +1,7 @@
 module Hadolint.Formatter.Format
   ( severityText,
     stripNewlines,
+    errorMessage,
     errorMessageLine,
     errorPosition,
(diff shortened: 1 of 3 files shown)


lorenzo:
Great job!

PR #591, opened 2021-04-03
DL3059: Multiple consecutive `RUN` instructions

m-ildefons:
- Add new rule linting for multiple consecutive run instructions that
  could be consolidated.
- Add tests for that rule.

On of the first bullet points in the Docker best practices is that one
should merge multiple `RUN` instructions into one, where possible to
reduce the layer count:
https://docs.docker.com/develop/dev-best-practices/
This rule informs with a severity of `DLInfoC`, where this is possible.

### How to verify it
The following should trigger rule DL3059:
```Dockerfile
RUN /foo.sh
RUN /bar.sh
```
None of these should trigger rule DL3059:
```Dockerfile
RUN /foo.sh
WORKDIR /
RUN /bar.sh
```
```Dockerfile
RUN /foo.sh \
 && /bar.sh
```

### Notes
I have omitted `onBuildRuleCatches[Not]` unit tests, because they don't behave properly. In particular, this one:
```Haskell
onBuildRuleCatchesNot "DL3059" "RUN /foo.sh\nWORKDIR /\nRUN /bar.sh
```
This is because the `onBuildRuleCatchesNot` only wraps `RUN` instructions and no other instructions in an `ONBUILD` context. This is probably wrong as other instructions (e.g. `LABEL`, `COPY`, `ADD`...) are allowed as well in an `ONBUILD` context.

As usual, I'll add a wiki page once this rule makes it into `master`.

PR #595, opened 2021-04-07
DL3059: improve behaviour when considering flags

m-ildefons:
Refine behaviour of DL3059 when dealing with `RUN` flags. These flags
are Docker buildkit specific, but may significantly alter behaviour of
`RUN` commands.

- Do warn when two consecutive `RUN` instructions are found with the
  same flags
- Do not warn when two consecutive `RUN` instructions have differen
  flags

As explained in [this discussion](https://github.com/hadolint/hadolint/pull/591#issuecomment-814595691) consolidation of consecutive `RUN` instructions may be undesired, if there are different flags for each of the instructions. In addition to that, consolidation may not be possible, if both `RUN` instructions have flags, but the flags are different.

### How to verify it
These Dockerfile should no longer trigger DL3059:
```Dockerfile
RUN --mount=type=secret,id=foo /foo.sh
RUN /bar.sh
```
```Dockerfile
RUN --mount=type=secret,id=foo /foo.sh
RUN --mount=type=secret,id=bar /bar.sh
```

The following Dockerfiles should still trigger DL3059:
```Dockerfile
RUN --mount=type=secret,id=foo /foo.sh
RUN --mount=type=secret,id=foo /bar.sh
```
```Dockerfile
RUN /foo.sh
RUN /bar.sh
```

### DIFF
--- a/src/Hadolint/Rule/DL3059.hs
+++ b/src/Hadolint/Rule/DL3059.hs
@@ -4,23 +4,26 @@ import Hadolint.Rule
 import Language.Docker.Syntax
 
 
-type Acc = Int
+data Acc
+  = Acc RunFlags
+  | Empty
+  deriving (Eq, Show)
 
 rule :: Rule args
-rule = customRule check (emptyState 0)
+rule = customRule check (emptyState Empty)
   where
     code = "DL3059"
     severity = DLInfoC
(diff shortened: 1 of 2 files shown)


lorenzo:
looks great, thanks!

PR #596, opened 2021-04-09
DL3005: allow apt-get upgrade (fix #562)

### LINKED ISSUE #562: DL3005 is wrong: security updates should be in most Dockerfiles
https://github.com/hadolint/hadolint/wiki/DL3005 suggests not running package updates. This is a bad idea, because it means many users won't get security updates installed.

For example, the official CentOS images only get sregenerated every few months. At time of filing this issue, the `centos:8` image hasn't been updated for 3 months (https://hub.docker.com/_/centos?tab=tags&page=1&ordering=last_updated). (Yes, CentOS is EOLing soon, but that's irrelevant to this point).

Similarly, while official Python Docker do get rebuilt more often, there are often windows of a few days where Debian has released security updates, but images haven't been regenerated. Without `apt-get upgrade`, those updates will not get installed in a timely manner.

stek29:
Closes #562

### What I did
Removed apt-get upgrade from DL3005, only dist-upgrade is checked

### How I did it
Updated README, rule and tests for that rule

### How to verify it
list a docker file with `RUN apt-get update && apt-get upgrade` in it

note: [Wiki](https://github.com/hadolint/hadolint/wiki/DL3005) still has to be updated


lorenzo:
thanks!

PR #600, opened 2021-04-13
Update Config.hs

Mazorius:
### What I did

From documentation the key should not `allowedRegistries`. Changed it too `trustedRegistries`.

### How I did it

Change the error output.

### How to verify it

—


lorenzo:
thanks!

Issue #585, opened 2021-03-29
DL3042 does not respect PIP_NO_CACHE_DIR


slsyy:
- [x] This is a bug report
- [ ] This is a feature request
- [x] I searched existing issues before opening this one

### Expected behavior

No error

### Actual behavior

```
Dockerfile:54 DL3042 warning: Avoid use of cache directory with pip. Use `pip install --no-cache-dir <package>
```

### Steps to reproduce the behavior

**Dockerfile** (if relevant)

```Dockerfile
ENV PIP_NO_CACHE_DIR=1
RUN pip3 install some_package
```

The `PIP_NO_CACHE_DIR=1` is read by the `pip`, and the behavior is the same, as I would use a `--no-cache-dir` flag. I need to use that variable, because in my `Dockerfile` I use a script, which call the `pip` internally, and the env variable is the only way to disable a caching

lorenzo:
good idea! would you like to contribute this feature?

PR #604, opened 2021-04-16
DL3042: respect `PIP_NO_CACHE_DIR` variable

m-ildefons:
Make DL3042 respect `PIP_NO_CACHE_DIR` variable, when it is set with an
`ENV` instruction.

fixes: #585

### Notes
~This fix only applies when `PIP_NO_CACHE_DIR` is set via the `ENV` instruction. This is because Hadolint does not expose the whole Shellcheck AST internally, so when parsing something like `RUN export PIP_NO_CACHE_DIR=1`, the export command is exposed, but the environment variable is not - at least not easily.~
 Fixed the recognition of `PIP_NO_CACHE_DIR` within the `RUN` instruction as well.

### How to verify it
```Dockerfile
RUN PIP_NO_CACHE_DIR=1 pip install wheel
RUN export PIP_NO_CACHE_DIR=1 && pip install numpy
ENV PIP_NO_CACHE_DIR=1
RUN pip install six
```
should no longer trigger DL3042.

### DIFF
--- a/src/Hadolint/Rule/DL3042.hs
+++ b/src/Hadolint/Rule/DL3042.hs
@@ -1,24 +1,65 @@
 module Hadolint.Rule.DL3042 (rule) where
 
 import Data.List (isInfixOf)
+import Data.Map.Strict (Map)
+import qualified Data.Map.Strict as Map
+import Data.Maybe (fromMaybe, isNothing, isJust, fromJust)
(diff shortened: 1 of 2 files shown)


lorenzo:
thanks, great work!

slsyy:
Thank you ❤️

Issue #607, opened 2021-04-19
Option to not fail based on severity


PedroArvela:
- [ ] This is a bug report
- [x] This is a feature request
- [x] I searched existing issues before opening this one

### Expected behavior

We have a CI pipeline that succeeds or fails based on hadolint's return code. Errors of type `info` and `style` should not cause hadolint to fail. We would like to be able to set what severities would cause the linter to fail.

### Actual behavior

Linter failure can only be globally disabled, not disabled by severity type. Ignores can only be added per individual codes, not by severity type.

### Steps to reproduce the behavior

Create a Dockerfile with an `info` level severity.

Run `hadolint Dockerfile`

The return code will be `1` instead of `0` even though the severity type of the error is `info`.

**Output of `hadolint --version` or
  `docker run --rm hadolint/hadolint hadolint --version` or
  `docker run --rm ghcr.io/hadolint/hadolint hadolint --version`:**

```bash
Haskell Dockerfile Linter v2.2.0-0-ga34e5f3
```

**Dockerfile** (if relevant)

```Dockerfile
FROM debian

# Example dockerfile with an info level severity lint error
RUN echo hello
RUN echo world
```

**Additional environment details (OS, stack version, etc.)**

lorenzo:
thanks, this is a great idea. If you want to help implement it, I can offer help and guidance

holgergp:
I'd like to do this, but I won't find the time to do this before this before the weekend. If someone has time before that, please go for it.
I have only basic Haskell knowledge though, I've never done anything beyond katas, so this might get a little rocky.
But it would be awesome to improve my skills here. 😀

lorenzo:
@holgergp happy to offer guidance, let me know if you have questions. I can come up with a high-level plan for you later today

Issue #602, opened 2021-04-13
Improve .hadolint.yaml documentation


Mazorius:
- [ ] This is a bug report
- [x] This is a feature request
- [x] I searched existing issues before opening this one

# Background

As seen in #600 the error message is not up to date.

Maybe you should update the [Hadolint JSON schema] as requested in #601 (https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/hadolint.json) and mention this in the error instead of printing examples?

Or a detailed configuration file description under documentation could be established which will link to the JSON schema.

What do you think about this?

lorenzo:
Sounds like a good idea. I'm all for improving the documentation. I like printing examples as help, though.

Would you like to take on this task?

Mazorius:
I will try 😄

lorenzo:
thanks!!

Mazorius:
@lorenzo do you want to stay with markdown or is AsciiDoc also possible?

lorenzo:
@markus1189 markdown.

m-ildefons:
I'm looking into updating the schema for schema store. There is the option to [self-host](https://github.com/SchemaStore/schemastore/blob/master/CONTRIBUTING.md#self-hosting-schemas) such a schema as part of this repository. Or I can just submit a PR there. Which is preferred?

Mazorius:
I think the best way is, to create a self-hosted and register always the latest release version in the catalog during a release?
Like for example angular is doing it.
What do you think @lorenzo ?

lorenzo:
yeah, that makes sense to me. Let me know if you need help doing that @Mazorius

PR #613, opened 2021-04-21
options: introduce failure severity threshold

m-ildefons:
- add new option `--threshold`/`-t` for specifying a failure threshold

The specified failure threshold (info by default) will cause hadolint to
not return with a failure code when only rules with a lesser severity
 are violated.
Accepted values are `error`, `warning`, `info`, `style`, `ignore` and
`none`, where `info` is the default and `ignore` and `none` both cause
Hadolint to exit with an error code if any rule is violated. A threshold
of level `error` has the same effect as passing `--no-fail`.

fixes: #607

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -1,8 +1,3 @@
-{-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE DeriveGeneric #-}
-{-# LANGUAGE NamedFieldPuns #-}
-{-# LANGUAGE TemplateHaskell #-}
-
 module Main where
 
 import Control.Applicative
@@ -18,6 +13,7 @@ import qualified Data.Version
 import qualified Development.GitRev
 import qualified Hadolint
 import qualified Hadolint.Rule as Rule
+import qualified Hadolint.Formatter.Format as Format
 import Options.Applicative
   ( Parser,
     action,
@@ -75,6 +71,15 @@ showFormat Hadolint.GitlabCodeclimateJson = "gitlab_codeclimate"
 showFormat Hadolint.Checkstyle = "checkstyle"
 showFormat Hadolint.Codacy = "codacy"
 
+toNofailSeverity :: String -> Maybe Rule.DLSeverity
+toNofailSeverity "error" = Just Rule.DLErrorC
+toNofailSeverity "warning" = Just Rule.DLWarningC
+toNofailSeverity "info" = Just Rule.DLInfoC
+toNofailSeverity "style" = Just Rule.DLStyleC
+toNofailSeverity "ignore" = Just Rule.DLIgnoreC
+toNofailSeverity "none" = Just Rule.DLIgnoreC
+toNofailSeverity _ = Nothing
+
 parseOptions :: Parser CommandOptions
 parseOptions =
   CommandOptions
@@ -89,6 +94,20 @@ parseOptions =
     version = switch (long "version" <> short 'v' <> help "Show version")
 
     noFail = switch (long "no-fail" <> help "Don't exit with a failure status code when any rule is violated")
+
+    noFailCutoff =
+      option
+        (maybeReader toNofailSeverity)
+        ( short 't'
+            <> long "failure-theshold"
+            <> help "Exit with failure code only when rules with a severity \
+                    \above THRESHOLD are violated. Accepted values: \
+                    \[error | warning | info | style | ignore | none]"
+            <> value Rule.DLInfoC
+            <> metavar "THRESHOLD"
+            <> showDefaultWith (Text.unpack . Format.severityText)
+            <> completeWith ["error", "warning", "info", "style", "ignore", "none"]
+        )
 
     nocolor = switch (long "no-color" <> help "Don't colorize output")
 
@@ -170,6 +189,7 @@ parseOptions =
         <*> styleList
         <*> ignoreList
         <*> parseRulesConfig
+        <*> noFailCutoff
 
     labels = Map.fromList
         <$> many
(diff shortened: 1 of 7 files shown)


app/Main.hs
```diff
@@ -75,11 +77,21 @@ showFormat Hadolint.GitlabCodeclimateJson = "gitlab_codeclimate"
 showFormat Hadolint.Checkstyle = "checkstyle"
 showFormat Hadolint.Codacy = "codacy"
 
+toNofailSeverity :: String -> Maybe Rule.DLSeverity
+toNofailSeverity "error" = Just Rule.DLErrorC
+toNofailSeverity "warning" = Just Rule.DLWarningC
+toNofailSeverity "info" = Just Rule.DLInfoC
+toNofailSeverity "style" = Just Rule.DLStyleC
+toNofailSeverity "ignore" = Just Rule.DLIgnoreC
+toNofailSeverity "none" = Just Rule.DLIgnoreC
```
lorenzo, line 86: Why do we need the none value?

lorenzo (commented):
I'm wondering if this option should be part of the configuration file as well

app/Main.hs
m-ildefons: While `ignore` fits well within the context of the severity, I think it is terrible user experience and very unintuitive. If you don't share this opinion, I'll remove it and add documentation accordingly.

m-ildefons:
I can make it part of the configuration.

app/Main.hs
lorenzo: makes sense to me, you can leave it

lorenzo:
> I've taken the opportunity to add the schemastore.org schema to the repo in order to have it self-hosted in the future, as discussed in #602.

Sounds awesome!

> I'm not sure anymore if I should rename the command line option to --failure-threshold

I like the new name and consistency, feel free to change it.

> I've tried the Haskell language server you recommended and it's very nice

Yeah, it's kind of life-changing haha

>  sometimes not clear old markers when they get resolved

I'm not experiencing the same. I often try with a separate set of vim configuration files from scratch periodically to make sure I'm not accumulating old cruft in my config files.

contrib/hadolint.json
```diff
@@ -0,0 +1,598 @@
...
+            "description": "var is referenced but not assigned.",
+            "$comment": "https://github.com/koalaman/shellcheck/wiki/SC2154"
+          },
+          {
+            "const": "SC2155",
+            "description": "Declare and assign separately to avoid masking return values.",
+            "$comment": "https://github.com/koalaman/shellcheck/wiki/SC2155"
+          },
+          {
+            "const": "SC2164",
```
lorenzo, line 583: Does this list contains all of the shellcheck possible errors?

m-ildefons: No. Just the ones from the readme. They are inherited from the previous iteration from schemastore.org.
I wonder if they should be in here at all.

harlowja:
Nice, looking forward to this change.

lorenzo:
Great work!

app/Main.hs
```diff
@@ -90,6 +95,20 @@ parseOptions =
 
     noFail = switch (long "no-fail" <> help "Don't exit with a failure status code when any rule is violated")
 
+    noFailCutoff =
+      option
+        (maybeReader toNofailSeverity)
+        ( short 't'
+            <> long "failure-theshold"
+            <> help "Exit with failure code only when rules with a severity \
+                    \above THRESHOLD are violated. Accepted values: \
```
n0v1, line 104: According to my tests setting THRESHOLD to `info` will cause exit code 1 when rules with a severity above **or equal** to `info` are violated. So I'd suggest to change to docs to:
```
            <> help "Exit with failure code only when rules with a severity \
                    \equal to or above THRESHOLD are violated. Accepted values:\
                    \[error | warning | info | style | ignore | none]"
```

n0v1: See https://github.com/hadolint/hadolint/pull/639

m-ildefons: Thanks for the bug report and the PR. But could you please detail how you determined this, since there was a typo in the long option until today and I can't reproduce the finding on the current master branch:
```
│ ~/repos/hadolint │ᐅ hadolint --failure-threshold=info ~/tmp/Dockerfile
/home/moritz/tmp/Dockerfile:3 DL3059 info: Multiple consecutive `RUN` instructions. Consider consolidation.
│ ~/repos/hadolint │ᐅ echo $?
0
│ ~/repos/hadolint │ᐅ 
(master) / 3aba485 11 hours ago
```
with Dockerfile
```Dockerfile
FROM scratch
RUN a
RUN b
```
Unfortunately, we currently don't have integration tests set up for this.

n0v1: I used the docker container with version 2.4.0 (and typo) for testing this. The Dockerfile is identical to yours.

```cmd
$ docker run --rm -it -v %cd%/Dockerfile:/tmp/Dockerfile hadolint/hadolint:2.4.0 /bin/hadolint --failure-theshold info /tmp/Dockerfile & echo %errorlevel%
/tmp/Dockerfile:4 DL3059 info: Multiple consecutive `RUN` instructions. Consider consolidation.
1

$ docker run --rm -it -v %cd%/Dockerfile:/tmp/Dockerfile hadolint/hadolint:2.4.0 /bin/hadolint --failure-theshold warning /tmp/Dockerfile & echo %errorlevel%
/tmp/Dockerfile:4 DL3059 info: Multiple consecutive `RUN` instructions. Consider consolidation.
0
```

Checking again this is not always reproducible for me. When first executing hadolint with `--failure-theshold ignore` (exit code 1) and then with `--failure-theshold info` it exits with code 1 at first. When running with `info` again it gives exit code 0. Am I doing something wrong or is this an issue with the Windows command line?

I can not reproduce this strange behaviour in WSL Ubuntu (or Powershell), so this is probably not caused by hadolint.
```bash
$ docker run --rm -it -v $(pwd)/Dockerfile:/tmp/Dockerfile hadolint/hadolint:2.4.0 /bin/hadolint --failure-theshold info /tmp/Dockerfile; ec
ho $?
```

n0v1: After sleeping on it for a night, IMHO the current implementation is not ideal from a usability perspective: When I want hadolint to fail on `info` rules, I would have to consult the severity list and lookup the next lower level (which is `style`). Also the word `failure` in `--failure-threshold` implies to me that hadolint would fail when a rule with the given severity is violated.

I'd suggest du adjust the implementation to fail on the given severity and change the default value of `--failure-threshold` from `info` to `warning`. What do you think @m-ildefons, @lorenzo?

m-ildefons: I deliberately chose the current behaviour, because this let me reuse the rule severities that we already have and still map all necessary behaviours: from never failing to failing on any rule violation. But your are technically correct that the dictionary meaning of the word *threshold* implies that the change in behaviour occurs *at* and not just above that level.

With https://github.com/hadolint/hadolint/pull/639 merged, now program behaviour is not reflected in the docs anymore.
I'm still reluctant to change this now because of the implication to current users, whose CI pipelines might break. I'd like to defer this decision to @lorenzo .

PR #615, opened 2021-04-26
feat(main): log configuration file used

### LINKED ISSUE #610: Log which config file was used
- [ ] This is a bug report

It would be helpful if hadolint output a log message indicating which config file it was using. Right now it doesn't say anything, so you can't easily tell. For example, if you accidentally name the file `hadolint.yaml` instead of `.hadolint.yaml`, you won't know unless you explicitly test a negative case.

seaerchin:
## Problem 
Users want to know what is the configuration file used so that troubleshooting in the event of errors can be made easier
Closes #610 

## Solution
This PR adds a flag that, when set, displays the configuration file used.

## Alternatives Considered
1. always printing the configuration file at invocation of `hadolint`. this was decided against because this should be an opt in feature as there might be a significant majority of users using `hadolint` as is. 

### Manual Tests
1. invoke `hadolint` with the `-h` option. verify that the show config option is there.
2. invoke `hadolint` with `-V` or `--verbose` but without a configuration file specified. `hadolint` should work but print a message notifying the user that there is no configuration file found.
3. invoke `hadolint` with `-V` or `--verbose` with a valid configuration file specified. `hadolint` should work and print the configuration file path. 

**Note**: there are currently no unit tests. would appreciate it if someone could point out to me where to add it because `ConfigSpec` seems to be on the config file itself and i'm not too sure if it belongs in `Spec`

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -49,6 +49,7 @@ import Options.Applicative
 import qualified Paths_hadolint as Meta
 import System.Environment
 import System.Exit (exitFailure, exitSuccess)
+import System.IO (hPutStrLn, stderr)
 
 data CommandOptions = CommandOptions
   { showVersion :: Bool,
@@ -278,7 +279,7 @@ main = do
     execute CommandOptions {dockerfiles = []} =
       putStrLn "Please provide a Dockerfile" >> exitFailure
     execute cmd = do
-      when (showConfigFile cmd) (putStrLn $ getFilePathDescription (configFile cmd))
+      when (showConfigFile cmd) (hPutStrLn stderr $ getFilePathDescription (configFile cmd))
       lintConfig <- Hadolint.applyConfig (configFile cmd) (lintingOptions cmd)
       let files = NonEmpty.fromList (dockerfiles cmd)
       case lintConfig of
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -1,6 +1,12 @@
+{-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE NamedFieldPuns #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TemplateHaskell #-}
+
 module Main where
 
 import Control.Applicative
+import Control.Monad (when)
 import qualified Data.Bifunctor as Bifunctor
 import qualified Data.List.NonEmpty as NonEmpty
 import qualified Data.Map as Map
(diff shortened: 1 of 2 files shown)


app/Main.hs
```diff
@@ -234,6 +245,7 @@ main = do
     execute CommandOptions {dockerfiles = []} =
       putStrLn "Please provide a Dockerfile" >> exitFailure
     execute cmd = do
+      when (showConfigFile cmd) (putStrLn $ getFilePathDescription (configFile cmd))
```
lorenzo, line 248: I think this should got to stderr, instead of stdout, so that the parsing of error messages is not disrupted

seaerchin: sure, this sounds good! could i just get extra clarification on the parsing of error messages and how it's being done right now?

lorenzo: Oh, that is being done by third parties, like editors and CI systems

seaerchin: Ah I see. Thank you

rittneje:
@lorenzo @seaerchin Most other tools I've used would output this type of information when you pass a `-v` (as in "verbose") flag. I would suggest doing that rather than making a flag specifically for logging the config file path. That way we can also log other useful information in the future based on that same flag.

seaerchin:
@rittneje @lorenzo that sounds like a good idea! however, i'm not sure what other information does `hadolint` have due to my lack of familiarity with the codebase. would you be amenable to me just changing the flag to be -v / --verbose and opening a subsequent PR for logging other useful information?

lorenzo:
sounds like a good idea!

lorenzo:
This branch now needs to be rebased

seaerchin:
@lorenzo no worries; could i just ask if the workflow for commits is to rebase or to merge? my main worry is that for rebases, you would have to review the whole PR again, which might not be an issue here because this PR is small but might be an issue for a future PR.

also, on a separate note, would you be fine with my suggestion of first creating a verbose option that solely outputs the config file name before adding in more descriptive information? my main reason for doing so is that i don't want scope creep to occur for this PR and due to my unfamiliarity with both haskell and the codebase, prevent any possible errors through limiting the scope

lorenzo:
Feel free to merge, if that is easier for you. I really don’t have a suggestion and I don’t mind reviewing again

lorenzo:
and yea, I’m ok with the idea of the verbose flag :)

seaerchin:
I'll rebase and request another review! But this should be seen as a logging of just the config file rather than as a verbose flag. I'll work on the actual verbose feature and raise it in another follow-up PR!

seaerchin:
i think i'll raise the verbose flag as another issue and actually, after thinking through it, we do have a few options. 

we could do something similar to npm, where we have a log file which we always write to and is enabled to `verbose` by default and in the event of an error, displays the name of the log file. 

another option is to just have it as what it is right now, and write the information to `stderr`.

lorenzo:
We could just always output the config file name to stderr if one was found. I don’t think that would be a problem to anyone

seaerchin:
@lorenzo fyi, updated some stuff! would appreciate if you could take a look!

lorenzo:
thank you, it looks great!

PR #617, opened 2021-04-28
refactor(config.hs): adds additional accepted config file

### LINKED ISSUE #614: Recognize .hadolint.yml along with .hadolint.yaml as configuration file
- [ ] This is a bug report

### Expected behavior

Recognize `.hadolint.yml`

### Actual behavior

Recognizes only `.hadolint.yaml` as per

https://github.com/hadolint/hadolint/blob/74f675eabf910fde339e01613c117c7b7065613b/src/Hadolint/Config.hs#L88-L90

------

It seems like a simple change, but I wish I was familiar with Haskell to offer a PR.

seaerchin:
### Problem
`.yaml` is [functionally identical](https://stackoverflow.com/questions/21059124/is-it-yaml-or-yml) to `.yml` and hence, users should be able to specify the config file in either format for `hadolint`.

This closes #614 

### Solution
This PR adds support for an additional config file `.hadolint.yml` for `hadolint` to use. Existing functionality was kept intact and `hadolint` should prefer the extension `.yaml` still over `.yml`. 

### Implementation Details
Changes `findConfig` from taking `FilePath` to a list of `FilePath`. Adds a constant called `acceptedConfigs` 

### Manual Tests 
- [ ] Create a file called `.hadolint.yml` in your current working directory. Invoke `hadolint` in the current working directory. This should call `hadolint` with the rules specified 
- [ ] Create a file called `.hadolint.yml` in `$XDG_CONFIG_HOME` (can `echo` to find out where this is on linux). Invoke `hadolint` in the current working directory. This should call `hadolint` with the rules specified.

### DIFF
--- a/src/Hadolint/Config.hs
+++ b/src/Hadolint/Config.hs
@@ -85,10 +85,12 @@ applyConfig maybeConfig o
       Nothing -> return (Right o)
       Just config -> parseAndApply config
   where
+    acceptedConfigs = [".hadolint.yaml", ".hadolint.yml"]
+
     findConfig = do
-      localConfigFile <- (</> ".hadolint.yaml") <$> getCurrentDirectory
-      configFile <- getXdgDirectory XdgConfig "hadolint.yaml"
-      listToMaybe <$> filterM doesFileExist [localConfigFile, configFile]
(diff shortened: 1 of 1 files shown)


lorenzo:
thanks @seaerchin :)

mloskot:
Thank you!

Issue #619, opened 2021-04-30
DL3060 unwarranted in all but the final stage of multi-stage Dockerfile builds


robdesbois:
- [x] This is a bug report
- [ ] This is a feature request
- [x] I searched existing issues before opening this one

### Expected behavior

I expected DL3060 warnings only for the final stage of multi-stage builds, as having larger earlier stages only affects the build cache, not the final image size.

### Actual behavior

DL3060 warning is flagged for non-final stages that run `yarn install` not followed by `yarn cache clear`.

### Steps to reproduce the behavior

Run `hadolint` on the `Dockerfile` below:

```bash
➜  hadolint --version
Haskell Dockerfile Linter 2.3.0-no-git
```

**Dockerfile** (if relevant)

```Dockerfile
## first stage
FROM node:lts-alpine as builder
WORKDIR /usr

COPY package.json tsconfig.json yarn.lock ./
RUN yarn install --frozen-lockfile

COPY src ./src/
RUN yarn build

## second stage
FROM node:lts-alpine as main
WORKDIR /

COPY /usr/package.json .
COPY --from=builder /usr/node_modules ./node_modules/
COPY --from=builder /usr/dist ./dist/

USER node
ENTRYPOINT ["node", "./dist/index.js"]
EXPOSE 9000
```

**Additional environment details (OS, stack version, etc.)**
Ubuntu 20.04 kernel 5.4.0-72-generic

PR #620, opened 2021-05-01
DL3060: stage aware logic

m-ildefons:
- Enable tests for DL3060
- Make failure logic stage aware, meaning that a `yarn install` without
  a `yarn cache clean` will not trigger DL3060 anymore in build stages,
  but will still trigger in the final image.

fixes: #619

### How to verify it
```Dockerfile
FROM node:lts-alpine
RUN yarn install
FROM node:lts-alpint
RUN foo
```
should no longer trigger DL3060, while
```Dockerfile
FROM node:lts-alpine as stage1
RUN yarn install
FROM stage1
RUN foo
```
still should trigger DL3060.

### DIFF
--- a/src/Hadolint/Rule/DL3060.hs
+++ b/src/Hadolint/Rule/DL3060.hs
@@ -1,27 +1,74 @@
 module Hadolint.Rule.DL3060 (rule) where
 
+import qualified Data.Map.Strict as Map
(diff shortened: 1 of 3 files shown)


lorenzo:
thanks!

robdesbois:
Thanks, that was superfast! :rocket:

Issue #624, opened 2021-05-04
Config issues


andrewzigerelli:
- [x] This is a bug report
- [ ] This is a feature request
- [x] I searched existing issues before opening this one

### Problem 
.hadolint.yaml appears to be ignored.
It appears to be loaded correctly, as it detects errors for malformed .yaml.

### Expected behavior
Ignored rules from .yaml file are actually ignored, in this example ignore DL3059.

### Actual behavior
DL3059 not ignored.

### Steps to reproduce the behavior

**Output of `hadolint Dockerfile` or
  `hadolint --config .hadolint.yaml`**

```bash 
Dockerfile:8 DL3059 info: Multiple consecutive `RUN` instructions. Consider consolidation.
Dockerfile:8 SC1091 info: Not following: File not included in mock.
```

**Dockerfile** (if relevant)

```Dockerfile
FROM ubuntu:20.04

SHELL ["/bin/bash", "-l", "-c", "-o", "pipefail"]

WORKDIR /root/

RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.38.0/install.sh | bash
RUN source /root/.nvm/nvm.sh && \
nvm install node
```

**.hadolint.yaml**
```yaml
ignored:
  - DL3059
```

**Additional environment details (OS, stack version, etc.)**
stack 2.5.1
I cloned from the git repo and installed from source.

### Another related issue
The README says I can automatically use ~/.config/hadolint.yaml, but it looks like 4b949a832eacad7f0065d94027df1208f441fd97 doesn't check this directory (I don't know much Haskell though). I tried by deleting the hadolint.yaml in the same directory as the Dockerfile, and forcing ~/.config/hadolint.yaml to be malformed. It's not loaded unless I manually pass it with -c option.

lorenzo:
you built hadolint from source? Asking since that commit is not in any release yet

andrewzigerelli:
> you built hadolint from source? Asking since that commit is not in any release yet

yes, sorry, I included it in the details now.

PR #627, opened 2021-05-05
config: fix parsing problems

m-ildefons:
Parsing the contents of the `failure-threshold` property was done
improperly and resulted in all results being discarded in case the
property was absent. The observed behaviour was seemingly the same as if
the config file had not been found, but in fact it was a problem of the
parser.

- Make all severity overrides more solid as types by implementing the
  to-be-overridden rules as rule codes instead of `Text.Text`s
- Make `DLSeverity` an instance of `Yaml.FromYAML`, `Semigroup` and
  `Monoid` to properly handle parsing from yaml files and combining
  multiple severities.

With `DLSeverity` being a more sophisticated type and implementing its
own yaml parser, parsing the config and combining it with the
configuration obtained from commandline options has much more consistent
behaviours with respect to the existing configuration properties.

fixes: #624

### How to verify it
Using the following config file and Dockerfile should ignore the error within the Dockerfile:
```yaml
ignored:
  - DL3059
```
```Dockerfile
FROM debian:buster
RUN foo
RUN bar
```

### DIFF
--- a/src/Hadolint/Lint.hs
+++ b/src/Hadolint/Lint.hs
@@ -24,13 +24,13 @@ import qualified Language.Docker as Docker
 import Language.Docker.Parser (DockerfileError, Error)
 import Language.Docker.Syntax (Dockerfile)
 
-type ErrorRule = Text
+type ErrorRule = Hadolint.Rule.RuleCode
 
-type WarningRule = Text
+type WarningRule = Hadolint.Rule.RuleCode
 
-type InfoRule = Text
+type InfoRule = Hadolint.Rule.RuleCode
 
-type StyleRule = Text
+type StyleRule = Hadolint.Rule.RuleCode
 
 type IgnoreRule = Hadolint.Rule.RuleCode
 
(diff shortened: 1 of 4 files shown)


src/Hadolint/Lint.hs
```diff
@@ -56,10 +56,10 @@ instance Semigroup LintOptions where
       (a4 <> b4)
       (a5 <> b5)
       (a6 <> b6)
-      (min a7 b7)
+      (a7 <> b7)
```
lorenzo, line 59: Interesting that in had a gut feeling about this in the previous PR that it would be better to implement monopod for the type

src/Hadolint/Rule.hs
```diff
@@ -21,7 +21,30 @@ data DLSeverity
...
   | DLIgnoreC
-  deriving (Show, Eq, Ord, Generic, NFData)
+  deriving (Show, Read, Eq, Ord, Generic, NFData)
+
+instance Yaml.FromYAML DLSeverity where
+  parseYAML = withSeverity pure
+
+withSeverity :: (DLSeverity -> Yaml.Parser a) -> Yaml.Node Yaml.Pos -> Yaml.Parser a
+withSeverity f v@(Yaml.Scalar _ (Yaml.SStr b)) =
+  case Hadolint.Rule.readSeverity b of
```
lorenzo, line 31: I find it odd that you are using a fully qualified call for a function in the same module. Is there a reason for this?

lorenzo:
thanks for fixing it!

src/Hadolint/Rule.hs
m-ildefons: Initially I wanted to name the function just `read`, which would have required the fully qualified call, but then I had to rename it because the `read` function already exists for the `LabelType`.
I can probably remove the fully qualified part here and also below in `withLabelType`, if I rename `read` to `readLabelType`.

I might do that in the future. There are also some very similar `veryCustomRule`s which whose state tracking could be unified.

PR #639, opened 2021-05-11
Fix docs on failure-threshold option

n0v1:
This is a follow-up on hadolint#607. The given threshold is inclusive. I.e. setting threshold to `info` will cause exit code 1 when rules with a severity above **or equal** to `info`  are violated. The docs should reflect that.

Update; My test that lead to this conclusion seems broken. See discussion in https://github.com/hadolint/hadolint/pull/613#pullrequestreview-656369769.


lorenzo:
thanks!

Issue #638, opened 2021-05-11
Fix `Index too large` crash on Windows


cordis-dev:
- [x] This is a bug report

### Expected behavior

Given this `Dockerfile`:
```
FROM scratch
COPY a b 
```

I expect latest version - `2.4.0` - to run without crashing.

### Actual behavior

Command `hadolint-Windows-x86_64.exe" "Dockerfile"` generated this error:

```
hadolint-Windows-x86_64.exe: Data.Text.Internal.Fusion.Common.index: Index too large
CallStack (from HasCallStack):
  error, called at libraries\text\Data\Text\Internal\Fusion\Common.hs:943:24 in text-1.2.3.2:Data.Text.Internal.Fusion.Common
```

Note: `v2.0.0` does not crash. `v2.1.0` and up *does* crash.

**Additional environment details (OS, stack version, etc.)**

Windows 10 Pro (20H2)

PR #640, opened 2021-05-11
bugfix: avoid index overrun in DL3045

m-ildefons:
- fix index overrun in DL3045 with single character paths when checking
  for windows type paths.

fixes: #638

### DIFF
--- a/src/Hadolint/Rule/DL3045.hs
+++ b/src/Hadolint/Rule/DL3045.hs
@@ -85,5 +85,7 @@ quotePredicate c
 
 isWindowsAbsolute :: Text.Text -> Bool
 isWindowsAbsolute path
-  | Char.isLetter (Text.index path 0) && (':' == Text.index path 1) = True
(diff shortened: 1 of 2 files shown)


src/Hadolint/Rule/DL3045.hs
```diff
@@ -85,5 +85,7 @@ quotePredicate c
 
 isWindowsAbsolute :: Text.Text -> Bool
 isWindowsAbsolute path
-  | Char.isLetter (Text.index path 0) && (':' == Text.index path 1) = True
+  | Text.length path >= 2
+    && Char.isLetter (Text.index path 0)
+    && (':' == Text.index path 1) = True
```
lorenzo, line 90: Is there a version of that function that returns a maybe? We should use that instead

m-ildefons: Not exactly, but `uncons` is close enough, since this is at the start of `path`.

Issue #628, opened 2021-05-05
Hadolint-docker pre-commit doesn't respect rev parameter


mathbunnyru:
- [x] This is a bug report
- [ ] This is a feature request
- [x] I searched existing issues before opening this one

Hadolint-docker pre-commit doesn't respect rev parameter and always uses the latest docker image.

### Expected behavior

To use the exact version from `rev`.

### Actual behavior

Docker pulls latest image no matter what I say in `rev`.

### Steps to reproduce the behavior

1. Create an empty git dir or use the existing one.
2. Add `.pre-commit-config.yaml` config:

  ```yaml
  ---
  repos:
    - repo: https://github.com/hadolint/hadolint.git
      rev: v2.1.0
      hooks:
        - id: hadolint-docker
  ```

3. Run `pre-commit install` (pre-commit should be installed).
4. Create and `git add` a simple Dockerfile:

  ```Dockerfile
  FROM ubuntu:focal
  
  RUN ls
  RUN ls
  ```
5. Run `pre-commit run --all-files`.

You will see an error: `Dockerfile:4 DL3059 info: Multiple consecutive `RUN` instructions. Consider consolidation.`
Not, that I used `rev: v2.1.0` in my config.
And this error appeared only in v2.2.0 (https://github.com/hadolint/hadolint/releases/tag/v2.2.0)

Also, when you run the command for the first time, you'll see, that the `latest` tag is pulled from docker.

**Output of `docker run --rm hadolint/hadolint hadolint --version`

```
docker run --rm hadolint/hadolint hadolint --version
Haskell Dockerfile Linter v2.3.0-0-g96e61f8
```

mathbunnyru:
A little bit of investigation:
https://pre-commit.com/index.html#docker_image

We're using this form of running:
```yaml
-   id: dockerfile-no-entrypoint-2
    name: ...
    language: docker_image
    entry: my.registry.example.com/docker-image-3:latest my-exe
```

Note, that example explicitly tells `latest`, which is ok for an example.

The config in this repo doesn't pin version at all, which why latest is used..
https://github.com/hadolint/hadolint/blob/1231d47d6dda198bc95112d62b720e75ddd51ef3/.pre-commit-hooks.yaml#L7

mathbunnyru:
So, the simple solution would be to change from `hadolint/hadolint` to something like `hadolint/hadolint:v2.3.0` and update this before release.
I'm not sure that's ok in this repo (will have to update manually each time before release).

[...]

PR #647, opened 2021-05-22
Update docker version in pre-commit-hooks

mathbunnyru:
Related: https://github.com/hadolint/hadolint/issues/628

Please, next time update this before creating a release.
This way the release (basically, zip file), will get the same docker version as the release and it will be consistent.


mathbunnyru:
So, when releasing v2.4.2, you have to update this file before creating a new release.

Issue #651, opened 2021-06-03
--failure-threshold does not work as expected (exits 0) if set to level: 'error'


Dentrax:
- [x] This is a bug report
- [ ] This is a feature request
- [ ] I searched existing issues before opening this one

### Expected behavior

It should exit with `1`.

### Actual behavior

It exits with `0`.

### Steps to reproduce the behavior

1. `$ hadolint --failure-threshold error --verbose ./Dockerfile`
2. `echo $?  # prints 0`

```
No configuration was specified. Using default configuration
./Dockerfile:8 DL3020 error: Use COPY instead of ADD for files and folders
./Dockerfile:10 DL3000 error: Use absolute WORKDIR
./Dockerfile:19 DL3008 warning: Pin versions in apt get install. Instead of `apt-get install <package>` use `apt-get install <package>=<version>`
...
```

**Output of `hadolint --version`:**

```bash
Haskell Dockerfile Linter 2.4.1-no-git
```

druwadi:
I'm having the same issue with this, I changed the trusted registry to create an error, and the error still passed the threshold.

m-ildefons:
Sorry for the confusion on this. There was a lengthy discussion on whether or not the behavior should be inclusive or exclusive of the threshold itself [here](https://github.com/hadolint/hadolint/pull/613#discussion_r629914696) and the docs were updated [here](https://github.com/hadolint/hadolint/pull/639), but the behavior itself never changed.

oliv3r:
@m-ildefons This seems to be true, except for the `style` rule it seems. If one wants to trigger on style, this doesn't seem to be possible with `failure-threshold`?

PR #653, opened 2021-06-10
threshold: tune behaviour to match help message

m-ildefons:
The help message suggests that the failure threshold makes Hadolint exit
with failure when a rule above **or equal to** the threshold is
violated. This was introduced to the help message with
https://github.com/hadolint/hadolint/pull/639/commits/3121fe9d79e609979f7c5bfe5704ba22c5af7bc6
But never updated the behaviour to match the help message.

fixes: #651


lorenzo:
Thanks for fixing it!

Issue #644, opened 2021-05-18
Ignore shellcheck when using cmd as SHELL


okainov:
- [ ] This is a bug report
- [x] This is a feature request
- [x] I searched existing issues before opening this one

Kind of the same as  #246 but for `cmd`. Probably also for `powershell` as well.

See also #431.

### Expected behavior

When using `SHELL ["cmd.exe", "/s", "/c"]`, the commands in a RUN should not be run through ShellCheck

Issue #431, opened 2020-06-02
Support for Windows containers using `SHELL ["powershell"]`?


petemounce:
First of all, please let me say I deeply appreciate the time and effort you have put into this excellent tool; I use it for all my real-computer containers. I just tried it on something a bit different, a Windows container. I'm not really sure whether this is a bug report or a feature request; I vaguely understand that hadolint depends on shellcheck. 

(I don't think it's particularly reasonable of me, a complete stranger, to expect support for a whole new shell as a drive-by issue!)

- [x] This is a bug report
- [x] This is a feature request
- [x] I searched existing issues before opening this one
    - I found https://github.com/hadolint/hadolint/issues/246 which looks like it works by noticing `SHELL ["pwsh"]`.

### Expected behavior

Should understand that powershell is a _completely reasonable language_ to use within a dockerfile and the world should recognise its brilliance (or, maybe, warn that I'm crazy?).

I'd be totally happy with the same behaviour as #246 does for `pwsh`.

### Actual behavior

Does not recognise my SHELL directive and cannot really figure out the rest, I think.

### Steps to reproduce the behavior

```
$ hadolint workflow/dockers/go-build.img/Dockerfile.windows
workflow/dockers/go-build.img/Dockerfile.windows:5 SC1008 This shebang was unrecognized. ShellCheck only supports sh/bash/dash/ksh. Add a 'shell' directive to specify.
workflow/dockers/go-build.img/Dockerfile.windows:5 SC1073 Couldn't parse this function. Fix to allow more checks.
workflow/dockers/go-build.img/Dockerfile.windows:5 SC1065 Trying to declare parameters? Don't. Use () and refer to params as $1, $2..
workflow/dockers/go-build.img/Dockerfile.windows:5 SC1064 Expected a { to open the function definition.
workflow/dockers/go-build.img/Dockerfile.windows:5 SC1072  Fix any mentioned problems and try again.
workflow/dockers/go-build.img/Dockerfile.windows:9 SC1008 This shebang was unrecognized. ShellCheck only supports sh/bash/dash/ksh. Add a 'shell' directive to specify.
workflow/dockers/go-build.img/Dockerfile.windows:9 SC2239 Ensure the shebang uses an absolute path to the interpreter.
```

**Output of `hadolint --version` or
  `docker run --rm hadolint/hadolint hadolint --version`:**

```bash
$ hadolint --version
Haskell Dockerfile Linter v1.17.3-1-g3fab409
```

[...]

PR #654, opened 2021-06-10
shellcheck: ignore non-posix shells

m-ildefons:
- Ignore execution of shellcheck rules and DL4006 for non-posix shells
  other than pwsh as well, where said rules are not applicable.

fixes: #644 #431

### What I did

Expand the list of shells known to break with shellcheck and avoid execution of shellcheck when they are set as `SHELL`.

### How to verify it
There are tests for the relevant rules (Shellcheck and DL4006), but also this Dockerfile for example should no longer be complained about:
```Dockerfile
FROM mcr.microsoft.com/powershell:ubuntu-16.04
SHELL [ "cmd", "-c" ]
RUN Get-Variable PSVersionTable | Select-Object -ExpandProperty Value
```


lorenzo:
Thanks!

PR #631, opened 2021-05-06
Fix hadolint docker image in .pre-commit-hooks.yaml

mathbunnyru:
fixes https://github.com/hadolint/hadolint/issues/628

### What I did

The version of docker image is fixed in .pre-commit-hooks.yaml

### How I did it

Just added tag

### How to verify it

```
---
repos:
  - repo: https://github.com/mathbunnyru/hadolint.git
    rev: asalikhov/pre_commit_docker_version
    hooks:
      - id: hadolint-docker
```

If I prune docker images, on the first run it's easy to see, that exact version is pulled from docker:
`docker image rm hadolint/hadolint:v2.3.0`
`pre-commit run --all-files --verbose`

```
Unable to find image 'hadolint/hadolint:v2.3.0' locally
v2.3.0: Pulling from hadolint/hadolint
4fc54f9b225a: Pull complete
Digest: sha256:5bd624ce29f153036a3b03083af66f9ac040d2cb0673b2b4785394425e66f10b
Status: Image is up to date for hadolint/hadolint:v2.3.0
Unable to find image 'hadolint/hadolint:v2.3.0' locally
v2.3.0: Pulling from hadolint/hadolint
4fc54f9b225a: Pull complete
Digest: sha256:5bd624ce29f153036a3b03083af66f9ac040d2cb0673b2b4785394425e66f10b
Status: Image is up to date for hadolint/hadolint:v2.3.0
Unable to find image 'hadolint/hadolint:v2.3.0' locally
v2.3.0: Pulling from hadolint/hadolint
4fc54f9b225a: Pull complete
Digest: sha256:5bd624ce29f153036a3b03083af66f9ac040d2cb0673b2b4785394425e66f10b
Status: Downloaded newer image for hadolint/hadolint:v2.3.0
```

PR #657, opened 2021-06-19
Update .pre-commit-hooks.yaml

brl0:
Just a minor change updating the docker image tag to the latest version as mentioned in #631 and #647.

@asottile, in pre-commit is there a better way to indicate the version of the docker image to be pulled such that issues like #628 (not respecting rev, always pulling latest) are avoided but do not require manually updating the version in the `.pre-commit-hooks.yaml` file for each release?  If not currently, I would be happy to open an issue for that feature to be considered for pre-commit.

Thanks to all for the great tools!  Y'all are making people's lives better.


lorenzo:
Thanks!

Issue #662, opened 2021-06-23
 DL3018 warning: Pin versions in apk add. Instead of `apk add <package>` use `apk add <package>=<version>`


andrei-galkin:
- [x] This is a bug report
- [ ] This is a feature request
- [x] I searched existing issues before opening this one

### Expected behavior
hadolint reports no errors

### Actual behavior
hadolint reports:

/dev/stdin:3 DL3018 Pin versions in apk add. Instead of `apk add <package>` use `apk add <package>=<version>`
### Steps to reproduce the behavior
The installation of https://github.com/sgerrand/alpine-pkg-glibc cause the issue.

I use these steps to install alpine-pkg-glibc

FROM alpine:3.12
RUN curl -fSL https://alpine-pkgs.sgerrand.com/sgerrand.rsa.pub -o /etc/apk/keys/sgerrand.rsa.pub \
    && curl -fSL https://github.com/sgerrand/alpine-pkg-glibc/releases/download/2.32-r0/glibc-2.32-r0.apk -o glibc-2.32-r0.apk \
    && apk add glibc-2.32-r0.apk

**Output of `hadolint --version` or
  `docker run --rm hadolint/hadolint hadolint --version` or
  `docker run --rm ghcr.io/hadolint/hadolint hadolint --version`:**

Haskell Dockerfile Linter 2.5.0-no-git

PR #664, opened 2021-06-24
DL3018: Don't trigger when installing from .apk file  #662

m-ildefons:
DL3018 ensures version pinning. When installing from an .apk file, this
is pointless and should not trigger the rule, since it is equivalent to
version pinning.

fixes: #662

### How to verify it
This should no longer trigger DL3018:

```Dockerfile
RUN apk add mypackagefile-1.1.1.apk
```
Test case included.


lorenzo:
thanks!

Issue #411, opened 2020-04-09
build and release an arm64 version of hadolint


vielmetti:
- [ ] This is a bug report
- [x] This is a feature request
- [x] I searched existing issues before opening this one

Previously #314 noting failures.

### Expected behavior

Releases directory and docker image published to include a linux/arm64 version of `hadolint`.

### Actual behavior

Only linux/amd64 versions are available.

### Steps to reproduce the behavior

See https://hub.docker.com/r/hadolint/hadolint/tags

**Additional environment details (OS, stack version, etc.)**

Please note that substantial progress has been made on Haskell for aarch64 (arm64), and the report I have from my colleague who has gone through this motivates this work by noting that `hadolint` is part of the build process for several Kubernetes components, and that it can be built by using Haskell's published `stack` binaries.

lorenzo:
Is there any free CI service supporting arm64?

vielmetti:
Yes @lorenzo there is Travis CI - details about their progress here https://blog.travis-ci.com/arm-architecture-adoption-grows-on-travis-ci - as well as Drone Cloud https://blog.drone.io/continuous-integration-workloads-for-arm-grow-10x/

vielmetti:
some progress, after installing `stack` on arm64:

```
emv@01:~$ stack --version
Version 1.5.1 aarch64
```

```
emv@01:~/src/github.com/hadolint/hadolint$ stack install
Could not parse '/home/emv/src/github.com/hadolint/hadolint/stack.yaml':
Aeson exception:
Error in $['extra-deps'][0]: failed to parse field 'extra-deps': (Invalid package identifier: "ShellCheck-0.7.1@sha256:94a6ee5a38e2668204bc95fdc7539a9a4ca7230984157694409444530c23b5a4,3170","ShellCheck-0.7.1@sha256:94a6ee5a38e2668204bc95fdc7539a9a4ca7230984157694409444530c23b5a4,3170")
See http://docs.haskellstack.org/en/stable/yaml_configuration/
```

vielmetti:
I'll note that's a pretty old version of `stack`, but trying to upgrade it fails as follows:

```
emv@01:~/src/github.com/hadolint/hadolint$ stack upgrade --binary-only
Warning: Unknown value for architecture setting: "aarch64"
Binary upgrade not yet supported on arch: OtherArch "aarch64"
```

vielmetti:
[...]

PR #665, opened 2021-06-29
Build and release an arm64 version of hadolint docker image

morlay:
### What I did

may fix https://github.com/hadolint/hadolint/issues/411

### How I did it

* setup `stack build env` for arm64
* remove upx to avoid issure for arm64 builds
* updates github actions, then it could build multi-arch hadolint images  

### How to verify it

Demo: 

```bash
docker run --rm -i docker.io/morlay/hadolint:dev-arm64  < docker/Dockerfile 
```

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:stretch-slim AS builder
+FROM debian:stretch-slim AS stack-build-env-amd64
 
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -15,20 +15,46 @@ RUN apt-get update \
...
 RUN stack --no-terminal --install-ghc test --only-dependencies
 
-COPY . /opt/hadolint
 RUN scripts/fetch_version.sh \
   && stack install --ghc-options="-fPIC" --flag hadolint:static
 
-# COMPRESS WITH UPX
-RUN curl -sSL https://github.com/upx/upx/releases/download/v3.96/upx-3.96-amd64_linux.tar.xz \
-  | tar -x --xz --strip-components 1 upx-3.96-amd64_linux/upx \
-  && ./upx --best --ultra-brute /root/.local/bin/hadolint
```
lorenzo, line 29: Any reason why you removed this?

docker/Dockerfile
```diff
@@ -15,20 +15,46 @@ RUN apt-get update \
...
+    unxz /tmp/llvm.tar.xz && \
+    tar xfv /tmp/llvm.tar --strip-components 1 -C /usr && \
+    rm /tmp/llvm.tar
+
+RUN curl -sSL https://github.com/commercialhaskell/stack/releases/download/v2.7.1/stack-2.7.1-linux-$(uname -m).bin > /usr/local/bin/stack && \
+    chmod +x /usr/local/bin/stack
+
+FROM stack-build-env-${TARGETARCH} AS builder
+
+RUN stack setup ghc-8.10.4
```
lorenzo, line 44: I think this should be taken from a stack.yaml file instead (i.e. just call stack setup)

lorenzo:
this is great work, thanks for the contribution!

docker/Dockerfile
morlay: i got an error, when use upxed binary
`prot_exec|prot_write failed`

Not sure why. May because the build env in Ubuntu

morlay: i will try copy into debian as a new stage and enable upx tomorrow.

docker/Dockerfile
morlay: Sure. i will clean it. 

It is an example of arm64 build env from stack repo. 😂

I haven't use Haskell and toolchains before.

lorenzo: thanks again for adventuring!

docker/Dockerfile
morlay: `docker run --rm -i docker.io/morlay/hadolint:slim-arm64  < docker/Dockerfile`

upx in debian still have same errors.

so removing upx is safer

morlay:
@lorenzo 

i failed to compress with upx for arm64 version, always got error.

`prot_exec|prot_write failed`

lorenzo:
will this file now work for both platforms? (arm and x86)?

morlay:
yes.

docker buildx support it.

different arch use different stages

lorenzo:
Thank you so much for doing all this. I have merged your changes to master after adding a few modifications to the Dockerfile.

morlay:
@lorenzo is any plan to release a new version?

lorenzo:
I’ll have time to do it this weekend

morlay:
@lorenzo Thanks

Issue #671, opened 2021-07-16
False positive DL3019


jtdoepke:
- [ ] This is a bug report
- [x] This is a feature request
- [x] I searched existing issues before opening this one

### Expected behavior

Check `DL3019` shouldn't fire if a [Buildkit cache mount](https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md#run---mounttypecache) is used for the `/var/cache/apk` dir.

Related to #497.

### Actual behavior

```
-:3 DL3019 info: Use the `--no-cache` switch to avoid the need to use `--update` and remove `/var/cache/apk/*` when done installing packages
```

### Steps to reproduce the behavior

**Output of `hadolint --version` or
  `docker run --rm hadolint/hadolint hadolint --version` or
  `docker run --rm ghcr.io/hadolint/hadolint hadolint --version`:**

```bash
Haskell Dockerfile Linter v2.6.0-0-g7160964
```

**Dockerfile** (if relevant)

```Dockerfile
# syntax = docker/dockerfile:1.2
FROM alpine:3.14.0
RUN --mount=type=cache,target=/var/cache/apk apk add -U curl=7.77.0-r1
```

**Additional environment details (OS, stack version, etc.)**

N/A

m-ildefons:
Hi Jaye,
thanks for reporting this issue, DL3009 suffers from a similar issue as well. I've made a PR to fix those.
#497 unfortunately is a bit harder to fix because the cache directory changes depending on which user executes `pip`.

PR #672, opened 2021-07-20
Feature: Silence DL3009,DL3019 for cache mounts

m-ildefons:
In case the cache directories DL3009 and DL3019 respectively are
provided via BuildKit cache mount, the two rules need not warn a user to
remove the directories.

fixes: #671

### How to verify it
The following Dockerfiles should no longer emit warnings DL3009 or DL3019 respectively:
```Dockerfile
FROM alpine:3.14.0
RUN --mount=type=cache,target=/var/cache/apk apk add -U curl=7.77.0-r1
```

```Dockerfile
FROM debian:10
RUN --mount=type=cache,target=/var/lib/apt/lists/ apt-get update && apt-get install foo
```


lorenzo:
thanks!

PR #677, opened 2021-07-30
Allow upgrades by removing DL3017, DL3031, DL3039 and adjusting DL3035

realHorst:
### What I did
Removed the rules DL3017, DL3031, DL3039.
Adjusted rule DL3035 to allow upgrades, but not dist-upgrades.
### How I did it
Adjusted the rule DL3035 and its test.
Deleted the other rules and their tests.
Removed rules from `Spec.hs` and `Process.hs`.
Updated README, `hadolint.cabal` using hpack, and `contrib/hadolint.json`.
### Why i did it
The rationale of not allowing container upgrades inside containers is outdated (#562) and changes are already made to adjust other rules (#592) following the same advice.


lorenzo:
sorry for the late review and thank you so much for the contribution! I merged some things that caused a conflict. Would you mind fixing that?

realHorst:
Merged master and the format seems to have survived. Thanks for the review!

lorenzo:
thanks!

Issue #678, opened 2021-08-01
Document default severity of rules


phith0n:
- [ ] This is a bug report
- [x] This is a feature request
- [x] I searched existing issues before opening this one

Sometimes I have to adjust the rule severity through the `override` configure, but I don't know the current and default severity of a rule until I come across the error.

So I suggest adding the default severity to the document at `README.md` and wiki, such as:

| Rule                     | Severity                                    | Description                                                                                                                                         |
|:-------------------------------------------------------------|:-------------------------------------------------|:---------------------------------------------------------------------------------------------------|
| [DL3000](https://github.com/hadolint/hadolint/wiki/DL3000)   | error | Use absolute WORKDIR.                                                                                                                               |
| [DL3001](https://github.com/hadolint/hadolint/wiki/DL3001)   | info | For some bash commands it makes no sense running them in a Docker container like ssh, vim, shutdown, service, ps, free, top, kill, mount, ifconfig. |
| [DL3002](https://github.com/hadolint/hadolint/wiki/DL3002)   | warning | Last user should not be root.                                                                                                                       |

PR #680, opened 2021-08-03
documentation: List default severity

m-ildefons:
Add column documenting the default severity of each rule to the
readme.

fixes: #678

### Notes:
I have deliberately left out the default severities of the SCXXXX rules, because these are inherrited from ShellCheck and somewhat not under our control.


lorenzo:
thanks!

Issue #676, opened 2021-07-29
ADD does not extract zip archives, so DL3020 should not allow them.


kevinmehall:
- [x] This is a bug report
- [ ] This is a feature request
- [x] I searched existing issues before opening this one

The [DL3020](https://github.com/hadolint/hadolint/wiki/DL3020) (Use COPY instead of ADD for files and folders) rule is intended to allow `ADD` only for the purposes of extracting archives, recommending `COPY` when not doing so. It appears to explicitly allow filenames ending in `.zip`; however `ADD` [only supports tar archives](https://docs.docker.com/engine/reference/builder/#add) and does not support zip:

> If <src> is a local tar archive in a recognized compression format (identity, gzip, bzip2 or xz) then it is unpacked as a directory. 

Technically, it's the contents of the file that matter, not the extension, but the extension is a reasonable guess if you want to look at the Dockerfile only.

### Expected behavior

```
ADD foo.zip /foo/
```

should yield a DL3020 warning because `ADD` only supports tar archives, not zip, so the zip file will be copied into the image as-is, and `COPY` would work just as well.

### Actual behavior

No warning given when the extension is .zip

PR #681, opened 2021-08-03
DL3010, DL3020: fix edge-cases for archives

m-ildefons:
DL3020:
  - Don't warn that the user should use `ADD` instead of `COPY`, as this
    has the same result since zip files are not extracted.

DL3010:
  - Don't suggest using `ADD` when copying from a previous build stage,
    as `ADD` does not support copying from previous build stages.

General:
  - Unify the list of archive extensions to reduce data duplication.

fixes: #679
fixes: #676


lorenzo:
thanks!

PR #694, opened 2021-08-26
Add linux/aarch64 to Docker build workflow

ivy:
﻿

### What I did

This pull request builds on the work done in #665 to add support for shipping linux/aarch64 images (tested on my 13" Apple M1 MacBook Air). Apologies, I don't have much time to respond to feedback but I wanted to get a PR up to share with others.

### How I did it

- Extended the Docker workflow to build for x86_64 and aarch64 as separate jobs (to avoid timeouts) rather than the previous strategy of building within a single job.
- Extended the Dockerfile to manually install binary releases of Clang and Stack for aarch64.

The prior work from #665 used `ubuntu:20.04` as the base image. It should still be possible to continue with the current `debian:stretch-slim` image though that would likely require some adjustments to the Apt dependencies.

### How to verify it

I've tested using a linux/x86_64 build host (all green :+1:) and on darwin/aarch64 (native works 👍 / GHC under QEMU crashes :sob:).

```sh
docker build --platforms linux/aarch64 -f docker/Dockerfile .
docker build --platforms linux/amd64 -f docker/Dockerfile .
```

I didn't have time to test the changes to the workflow.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -11,14 +11,20 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        distro: ["scratch", "alpine", "debian"]
+        distro:
+          - scratch
+          - alpine
+          - debian
+        os:
+          - linux/amd64
+          - linux/aarch64
     steps:
       - uses: actions/checkout@v2
 
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v1
         with:
-          platforms: linux/amd64
+          platforms: ${{ matrix.os }}
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/docker.yml
```diff
@@ -11,14 +11,20 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        distro: ["scratch", "alpine", "debian"]
+        distro:
+          - scratch
+          - alpine
+          - debian
```
localheinz, line 17: What do you think about keeping entries sorted here alphabetically?

The order does not matter, so alphabetically makes the most sense.

.github/workflows/docker.yml
```diff
@@ -11,14 +11,20 @@ jobs:
...
+        os:
+          - linux/amd64
+          - linux/aarch64
```
localheinz, line 20: See above.

Issue #697, opened 2021-08-27
Pre-commit hook not updated


aballet:
- [x] This is a bug report
- [ ] This is a feature request
- [x] I searched existing issues before opening this one

### Expected behavior
When using v2.7.0 of the pre-commit hook of hadolint, it should use the tag v2.7.0 of the Docker image.

### Actual behavior
The pre-commit hook in version v2.7.0 tries to download the tag v2.6.1 in Docker Hub, which leads to an error since this specific version does not exist in Docker Hub.

### Steps to reproduce the behavior

Run `pre-commit install` then `pre-commit run --all-files` with this pre-commit configuration.

```bash
repos:
-   repo: https://github.com/hadolint/hadolint
    rev: v2.7.0
    hooks:
    -   id: hadolint-docker
```

The error is the following:

```bash
Unable to find image 'hadolint/hadolint:v2.6.1' locally
docker: Error response from daemon: manifest for hadolint/hadolint:v2.6.1 not found: manifest unknown: manifest unknown.
See 'docker run --help'.
```

PR #698, opened 2021-08-27
Fix version of Docker image in pre-commit hook

aballet:
This fixes issue #697.

### What I did
Updated the Docker version used in the pre-commit hook configuration file.


rally-carlos:
will there be a patch release for this?

PR #701, opened 2021-09-07
Drop quotes

m-ildefons:
### What I did
Generalize the bugfix in DL3020 from a few days ago. This really was about dropping quotes around a string and not about matching quotes at the beginning of a string.
Also bumping versions to incorporate the bugfix regarding heredoc termination in language-docker.

### How to verify it
This should parse fine now:
```Dockerfile
RUN <<EOF
echo $EOF
EOF
```
Tests for the quote dropping mechanic in DL3020 are included.

### DIFF
--- a/ThirdPartyNotices.txt
+++ b/ThirdPartyNotices.txt
@@ -1,4 +1,4 @@
-Package: hadolint-2.7.1 (License: GPL-3.0-only)
+Package: hadolint-2.7.2 (License: GPL-3.0-only)
 -----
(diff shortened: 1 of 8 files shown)


lorenzo:
@m-ildefons seems like it cannot build due to dependencies

m-ildefons:
Yeah, this includes a version bump of `language-docker` and therefore depends on https://github.com/hadolint/language-docker/pull/69 being merged and released. I really should have waited with this PR until after language-docker-10.1.2 is available via hackage.

PR #704, opened 2021-09-09
fomatters: Add SARIF formatter

m-ildefons:
Add formatter for SARIF output format. SARIF is the static analysis
results interchange format, a Microsoft/OASIS standard format for static
analysis tools.
See https://github.com/microsoft/sarif-tutorials
and https://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.html

Break long lines in `app/Main.hs`.

### How to verify it
`hadolint -f sarif Dockerfile` should generate a valid SARIF report for that dockerfile. Tools like [jsonschemavalidator](https://www.jsonschemavalidator.net/) can be used to validate the output against the schema spec.

Example dockerfile with output:
```Dockerfile
FROM debian:latest

WORKDIR ./foobar

RUN echo $foo
```
screengrab2

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -1,8 +1,3 @@
-{-# LANGUAGE DeriveGeneric #-}
-{-# LANGUAGE NamedFieldPuns #-}
-{-# LANGUAGE OverloadedStrings #-}
-{-# LANGUAGE TemplateHaskell #-}
-
 module Main where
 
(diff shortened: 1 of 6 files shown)


lorenzo:
thanks!

jbergstroem:
Thanks for this! Been looking at doing it myself

PR #707, opened 2021-09-15
DL3034: support --non-interactive and -n

gimler:
Whitelist `--non-interactive, -n` zypper install options.

from the zypper command help
```
--non-interactive, -n	Do not ask anything, use default answers
-y, --no-confirm            Don't require user interaction. Alias for the --non-interactive global
                            option.
```


lorenzo:
thanks so much!

Issue #709, opened 2021-09-18
Misinterpretation in DL3010


razamat:
- [x] This is a bug report
- [ ] This is a feature request
- [x] I searched existing issues before opening this one

### Expected behavior
In my Dockerfile, I want to copy *.tar.gz archive to a docker image. This archive will be later used by a *.sh script.
There is no reason for me to unpack the archive, I need it in packed form.
```COPY project/build/data/data.tar.gz /data/```

In my understanding, DL3010 should appear if hadolint would detect 
```COPY <archive_name>``` followed by a ```RUN tar -xf <same_archive_name>``` or something similar.

### Actual behavior
Using command ```COPY``` with a *.tar.gz file cause a complain about DL3010. 
Using command ```ADD``` as hadolint suggests, unpacks the archive.

There are a few possible workarounds like ```COPY /whole_directory/``` so hadolint doesn't notice the archive
or using ```ADD```  and "double packing" the file or simply ordering hadolint to ignore DL3010.

### Steps to reproduce the behavior
Use command COPY with a *.tar.gz file.

```bash
[2021-09-10T03:07:22.183Z] Dockerfile:53 DL3010 info: Use ADD for extracting archives into an image
```

**Dockerfile** (if relevant)

```Dockerfile
FROM [redacted]

RUN [redacted] \
 && mkdir -p /data

COPY project/build/script/ /bin/
COPY project/build/data/data.tar.gz /data/

CMD ["/bin/start.sh"]
```

m-ildefons:
Hi @razamat,
if you have good reason to ignore a particular rule, like in your case, you can also annotate that instruction to ignore that rule. E.g.:
```
[...]
# hadolint ignore=DL3010
COPY project/build/data/data.tar.gz /data/
[...]
```
Then Hadolint will omit that rule on that line. See [here](https://github.com/hadolint/hadolint#inline-ignores) for more information.
You can also disable rules entirely if you are sure that they will never be relevant to you.

razamat:
Thanks @m-ildefons,
I actually didn't know about "inline ignores", these might come handy in the future.

It's true that there are workarounds, "ignore" one and couple of others.
Nevertheless, I just found DL3010 to be very restrictive as copying an archive and not unpacking it can be a valid action.

m-ildefons:
Hi @razamat ,
[...]

PR #711, opened 2021-09-30
DL3010: Allow intentionally unextracted archives

m-ildefons:
- Add advanced logic to DL3010 to detect archive extraction.
- Add more specific tests for DL3010

Sometimes it is desired to copy an archive into a Docker image. In these
cases, DL3010 should not trigger. This can be detected by searching for
commands extracting an archive that was added by a `COPY` instruction
earlier.
The new behaviour triggers only if an archive is `COPY`ed into the image
_and_ extracted (e.g. by `RUN tar -xf archive-foo.tar.gz`).

fixes: #709

### DIFF
--- a/src/Hadolint/Rule/DL3010.hs
+++ b/src/Hadolint/Rule/DL3010.hs
@@ -1,21 +1,138 @@
 module Hadolint.Rule.DL3010 (rule) where
 
 import Data.Foldable (toList)
+import Data.Function ((&))
+import Data.List.NonEmpty (NonEmpty)
+import qualified Data.Set as Set
+import qualified Data.Sequence as Seq
 import qualified Data.Text as Text
 import Hadolint.Rule
+import qualified Hadolint.Shell as Shell
 import Language.Docker.Syntax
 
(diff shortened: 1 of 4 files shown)


src/Hadolint/Rule/DL3010.hs
```diff
@@ -1,21 +1,143 @@
...
+  Acc ->
+  Acc
+rememberArchives line paths target Empty =
+  if isArchive $ unTargetPath target
+    then Acc
+          { archives = Set.singleton (line, basename $ unTargetPath target),
+            extracted = Set.empty
+          }
+    else Acc
+          { archives =
```
lorenzo, line 104: I think these lines would benefit of using then `&` operator. Something like:

```haskell
paths
& toList
&  map (...)
& Set.fromList
...
```

m-ildefons: Yeah, that is much better. This operator is cool :-) I didn't know it existed. Thanks for telling me.

lorenzo:
great job!

PR #713, opened 2021-10-01
DL3059: RUNs and comments mixed

### LINKED ISSUE #712: DL3059 does not trigger if consecutive `RUN` instructions are separated by a comment
- [x] This is a bug report

(I guess this is somewhere in the middle of bug report and feature request. You decide).

### Expected behavior

Hadolint should find two consecutive `RUN` commands (even if there are comments in between them) and cite [DL3059](https://github.com/hadolint/hadolint/wiki/DL3059):

    /Dockerfile:5 DL3059 info: Multiple consecutive `RUN` instructions. Consider consolidation.

### Actual behavior

Because of the comment separating the `RUN` instructions the DL3059 is not triggered. I don't get any output and hadolint exits successfully (with statuscode `0`)

Run hadolint on the  following minimal-example Dockerfile below

```Dockerfile
FROM alpine:3.4

RUN apk update
# install my favorite editor
RUN apk add emacs
```

**Output of `hadolint --version`**

```bash
Haskell Dockerfile Linter v2.7.0-0-g3e13ec6
```

**Notes**

The code should be adopted here: https://github.com/hadolint/hadolint/blob/dcf946a769a831c86e7c7e234e67cbfc6abc9442/src/Hadolint/Rule/DL3059.hs#L19-L22
The problem occurs because `language-docker` does recognize `Comment` as an instruction: https://github.com/hadolint/language-docker/blob/a1e8ebf778f7b7999711e6c09ff537a0c2594345/src/Language/Docker/Syntax.hs#L357

Kaligule:
Fixes #712

### What I did

- added tests
- added code to DL3059 so that comments in the Dockerfile won't change the state (the state keeps track of the last line and if it was a `RUN` command.

### How I did it

Just added a case to the pattern matching, which makes the `otherwise` clause not get the comment-instructions.

### How to verify it

Run the tests and/or run hadolint on the Dockerfile from the issue.

### DIFF
--- a/src/Hadolint/Rule/DL3059.hs
+++ b/src/Hadolint/Rule/DL3059.hs
@@ -19,6 +19,7 @@ rule = customRule check (emptyState Empty)
     check line st (Run (RunArgs _ flags))
       | state st == Acc flags = st |> addFail CheckFailure {..}
       | otherwise = st |> modify (remember flags)
+    check _ st (Comment _) = st
     check _ st _ = st |> modify reset
 {-# INLINEABLE rule #-}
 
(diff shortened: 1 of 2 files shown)


lorenzo:
thanks so much!

Issue #691, opened 2021-08-23
Error when using podman with hadolint image


tschm:
- [x] This is a bug report
- [ ] This is a feature request
- [ ] I searched existing issues before opening this one

### Actual behavior
Hadolint hangs as the image doesn't seem to return anything when used with podman

### Steps to reproduce the behavior

Use any Dockerfile and

podman run --rm -i docker.io/hadolint/hadolint:2.6.0-debian < .

```

**Comment by tschm** [comment]
This may help: https://github.com/containers/podman/issues/3323

**Comment by wULLSnpAXbWZGYDYyhWTKKspEQoaYxXyhoisqHf** [comment]
> Use any Dockerfile and
> 
> podman run --rm -i docker.io/hadolint/hadolint:2.6.0-debian < .

This does not work on my Docker hosts either, meaning if I run your thing in Docker, it does not return anything, too, same as podman.

I believe what you might have meant to do with the command is to pass it just the Dockerfile instead of whole `.`, like so:
`podman run --rm -i docker.io/hadolint/hadolint:2.6.0-debian < Dockerfile`

For the record, the above works with *both* `Docker version 20.10.8, build 3967b7d28e` *and* `podman version 3.3.1`.

**Comment by 0uep** [comment]
Hi @tschm 

I also use Podman. The following script `hadolint` is in my `$PATH` and runs the current downloaded image `docker.io/hadolint/hadolint`:

```sh
#!/bin/bash
dockerfile="$1"
shift
podman run --rm -i docker.io/hadolint/hadolint hadolint "$@" - < "$dockerfile"
```

This script `hadolint` need to be used in this way:

`hadolint path/to/your/Dockerfile [OPTIONS]`

I sometimes upgrade my local image `docker.io/hadolint/hadolint`:

`podman pull docker.io/hadolint/hadolint`

I think this a good way to securely rum Hadolint with upgrade ease.

Moreover, this script works fine with the current version of the extension vscode-hadolint
https://github.com/michaellzc/vscode-hadolint

Note: I also use vscodium.com + open-vsx.org

0uep:
Hi @lorenzo 

Would you mind if the README explain how to use Hadolint with Podman?
You may also share the tips from my previous message.

Thank you for maintaining Hadolint. :smile:

lorenzo:
Your contribution will be very welcomed @0uep :)

PR #714, opened 2021-10-01
Add Podman example + Improve writing

0uep:
documents #691


lorenzo:
there's a small change that the linter is asking you to make. Can you fix that before I merge it @0uep ?

Thanks for the contribution :)

lorenzo:
thanks!

Issue #716, opened 2021-10-04
failure-threshold from config ignores 'style'


oliv3r:
- [x] This is a bug report
- [ ] This is a feature request
- [x] I searched existing issues before opening this one

### Expected behavior
The `failure-threshold` option should behave equally in a configuration file and from an argument.

### Actual behavior
`failure-threshold` seems to be only honored from the CLI, not the config file

### Steps to reproduce the behavior
* Create a `Dockerfile` with a style error in it; e.g. `LABEL Maintainer="a@b.com"` (Maintainer must be lower-caps)
* Create a `.hadolint.yaml` file `failure-threshold: "style"` in it.
* `docker run --rm --interactive hadolint/hadolint hadolint Dockerfile`
passes, due to style threshold not being failing
* `docker run --rm --interactive hadolint/hadolint hadolint --failure-threshold: "style" Dockerfile`
fails due to style error
*  `docker run --rm --interactive --volume "$(pwd):/workdir" -w "/workdir" hadolint/hadolint hadolint Dockerfile`
passes, as it ignores the failure threshold 'style' from the config file

Note, that other warnings work fine, e.g. 'upgrading' the style issue to an info issue, we do get a failure code.
```yaml
failure-threshold: "style"
override:
  info:
    - DL3048
```
Upgrading the threshold to "warning" then changes our exit behavior as expected.

**Output of `hadolint --version` or
  `docker run --rm hadolint/hadolint hadolint --version` or
  `docker run --rm ghcr.io/hadolint/hadolint hadolint --version`:**

```bash
Haskell Dockerfile Linter v2.7.0-0-g3e13ec6
```

m-ildefons:
Hi @oliv3r ,
thanks for the bug report. This is indeed a bug in the config file parsing logic. I'm working on a fix for this.
In the meantime, you can coerce Hadolint into your desired behavior by using the commandline options.

oliv3r:
@m-ildefons yeah; I think the bug-report showed that kind of ;) thanks for the tip anyway, much appreciated. As the linter is part of a template, and the local config file should be leading; that's not really possible; but the the wait for a fixed release is not a problem.

Issue #570, opened 2021-03-19
feat: configuration via env variables


lukasmrtvy:
- [ ] This is a bug report
- [X] This is a feature request
- [ ] I searched existing issues before opening this one

Would be nice to support configuration via ENV variables ( for example for --ignore `INGORE="DL3004,DL3005,DL3006"`), sometimes is not needed to create yaml config file ( mostly CI ), for dynamic configuration some templating must be done in hadonlint config file, then there is an alternative for suppressing rules from Dockerfile, last chance is creating some script for constructing dynamic arguments for hadolint, thats not easy in alpine image ( there is no bash )

m-ildefons:
I like the idea, but this can be done with a bit of shell trickery already. Note that the alpine image has no bash, but it does have `/bin/sh` from BusyBox v1.32.1.
```bash
$ docker run --rm -i -e HADOLINT_OPTIONS="--ignore=DL3033" hadolint/hadolint:latest-alpine /bin/sh -c "/bin/hadolint \$HADOLINT_OPTIONS -" < Dockerfile
```
There also is a debian based image, which comes with dash iirc. Beware that this only works in the alpine or debian based images, because those are the only one with a shell.

Or if you are not using Docker, it is much simpler:
```bash
$ HADOLINT_OPTIONS="--ignore=DL3004"
$ hadolint $HADOLINT_OPTIONS Dockerfile
```

lukasmrtvy:
@m-ildefons did You try to use multiple arguments ? `HADOLINT_OPTIONS="--ignore=DL3004 --ignore=DL3006"` this can not work. 

What is exact order of loading configuration ? 
1. arguments
2. hadolint config file 
3. dockerfile

? Thanks

m-ildefons:
Be sure to use correct quoting for the shell, multiple arguments work just fine this way
In order to use the config file it must obviously be mounted into the container, but works fine too. The pragmas from the Dockerfile are respected as well:
```
$ docker run \
      --rm \
      -i \
      -e HADOLINT_OPTIONS="--ignore=DL3000 --ignore=DL3001" \
      -v /path/to/hadolint.yaml:/hadolint.yaml \
      hadolint/hadolint:latest-alpine \
        /bin/sh -c "/bin/hadolint -c /hadolint.yaml \$HADOLINT_OPTIONS -" < Dockerfile
```
I've broken this command up into multiple lines for readability.

[...]

PR #721, opened 2021-10-17
Overhaul: Reorganize tests, redesign configuration

m-ildefons:
config: redesign configuration handling

- Redesign configuration handling to have a clear order of precedence
  default < environment < config file < command line options
- Implement more configuration options as environment variables and
  as options in the config files
- Define a clear default configuration
- Move configuration code into its own submodule and keep app/Main.hs
  clear of it
- Make configuration parsing and handling testable
- Add sensible locations to search for config files.
- Reorganize code and optimize for cleaner imports

tests: Use Hspec discovery to auto-discover tests

- Replace boilerplate code with hspec discovery.
  This also enables the use of Hspec hooks.
- Rename test submodules to conform with Hspecs expectations
- Make all fields in the config optional thus enabling proper priority
  based override behavior.
- Add tests for config parsing and configuration handling
- Add tests for output formats
- Add tests for command line argument parsing

other:

- Version bump language-docker 10.1.2 --> 10.2.0
- Verbose output now shows the whole running configuration
- Convert custom read functions for label types and severities to generic
  instance of `Read`.

fixes: #716
fixes: #570

This is a pretty big PR and should definitely be looked over with scrutiny before merging. I'm not quite sure yet if instantiating `Show` to produce human readable output is a good idea, or if I should rather use a proper pretty printing module. The code could also probably use some more unit tests around failure handling, especially for parsing configurations from the command line and environment variables. But it'd be nice to get some feedback now, so I know things are heading in the right direction.


lorenzo:
great job, as always!

radomirbosak:
Was the `trustedRegistries` config key renamed to `trusted-registries` here?

Issue #730, opened 2021-11-05
[2.8.0] pre-built binary releases for Linux and Mac report version as UNKNOWN


robert914:
- [x] This is a bug report
- [ ] This is a feature request
- [x] I searched existing issues before opening this one

The `-v`/`--version` output reports `UNKNOWN` for the pre-compiled release binaries distributed here: https://github.com/hadolint/hadolint/releases/tag/v2.8.0

### Expected behavior
On Mac:
```bash
❯ ./hadolint-Darwin-x86_64 -v
Haskell Dockerfile Linter 2.8.0-no-git
```
On Linux:
```bash
❯ ./hadolint-Linux-x86_64 -v
Haskell Dockerfile Linter 2.8.0-no-git
```

### Actual behavior
On Mac:
```bash
❯ ./hadolint-Darwin-x86_64 -v
Haskell Dockerfile Linter UNKNOWN
```
On Linux:
```bash
❯ ./hadolint-Linux-x86_64 -v
Haskell Dockerfile Linter UNKNOWN
```

### Steps to reproduce the behavior

See above.

**Dockerfile** (if relevant)

N/A

**Additional environment details (OS, stack version, etc.)**

Mac or Linux OS versions, possibly Windows (untested).

lorenzo:
no idea why, must be some change in the CI which I'm unaware of. Help appreciated

dmpe:
Jop, can also confirm that.

mjsteinbaugh:
I'm seeing this on macOS with Homebrew binary for v2.8.0.

darkvertex:
The Windows binary has this issue too.

PR #732, opened 2021-11-10
bugfix: display version string correctly

m-ildefons:
Construct the version string correctly when not compiling in a git
directory.
This bug was caused by a typo and a misunderstanding of the code when
the version string logic was moved from app/Main.hs to
src/Hadolint/Meta.hs in commit 3b1de5a

fixes: #730

### How to verify it
To verify that this actually fixes the issue, check out this commit, remove or rename the `.git` directory and compile Hadolint from source. The result should show the version string from the `.cabal` file and the `-no-git` ending.
```
$ hadolint -v
Haskell Dockerfile Linter 2.8.0-no-git
```
When compiling from source with the `.git` directory, the result should show the corresponding git tag and the short git hash.

### DIFF
--- a/src/Hadolint/Meta.hs
+++ b/src/Hadolint/Meta.hs
@@ -4,7 +4,9 @@ module Hadolint.Meta
   )
 where
 
-import qualified Development.GitRev
+import Data.Version (showVersion)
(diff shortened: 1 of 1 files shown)


lorenzo:
thanks for fixing it

Issue #747, opened 2021-12-16
Create a new flag that not allow the interpretation of "inline ignores"


iaghoCristian:
- [ ] This is a bug report
- [X] This is a feature request
- [X] I searched existing issues before opening this one

### Expected behavior
Possibility ignore the "inline hadolint ignores".
Create a flag that allows full interpretation of the Dockerfile, even if there are "inline ignores".

### Actual behavior
When using hadolint to scan Dockerfile is possible to bypass warnings by using "hadolint ignores" instead of actually fix the errors/warnings found in the Dockerfile

### Feature request specification:
**_--discard-inline-ignores_**  flag that does not allow the interpretation of "inline ignores"
e.g: ``` hadolint --discard-inline-ignores Dockerfile ```

felipecrs:
This is an interesting feature for us. We have a centrally managed hadolint configuration file which we use to lint Dockerfiles across many repositories, and we also don't want developers to ignore rules by their own.

Apart from the cli flag, it would also be cool if it could be set in the configuration file perhaps, like:

```yaml
failure-threshold: string
```

felipecrs:
OMG, this is brilliant! Thank you a lot for this, @m-ildefons!

PR #748, opened 2021-12-16
options: disable ignore pragma

m-ildefons:
- Add new configuration option allowing to disable inline ignore pragmas
- Add new rule DL1001 to allow warning on use of an inline ignore pragma

Inline ignore pragmas `# hadolint ignore=DLxxxx` can cause trouble in
organizations when it is not desirable that individual developers
circumvent centralized CI checks. This configuration options allows
(e.g. in such a CI pipeline) to just ignore the pragma and apply all
rules as usual. The option can be enabled via command line
`--disable-ignore-pragma`, environment variable
`HADOLINT_DISABLE_IGNORE_PRAGMA=1`, or config file
```yaml
disable-ignore-pragma=true
```

Just to prevent any confusion about the ignored pragma, the rule DL1001
can be used to fail the linting process regardless of how the dockerfile
fares, if desired. By default DL1001 is ignored but can be enabled by
setting its severity to the desired level (e.g. via command line option
`--warning=DL1001`).

Changes in the test suite:

- Add tests for `disable-ignore-pragma` options
- Add tests for rule DL1001
- Remove pointless configuration test for content of default
  configuration
- Change configuration tests to a more compact syntax relying on the
  default instance

By changing the manual construction of configuration values to utilize
the default instance, future changes to the configuration do not
necessitate many error prone changes to the test suite. This imporves
reliability and ease of use of the configuration test suite.

fixes: #747

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,8 @@ Available options:
                            format requirement `format`
   --strict-labels          Do not permit labels other than specified in
                            `label-schema`
+  --disable-ignore-pragma  Disable the inline ignore pragma `# hadolint
+                           ignore=DLxxxx` and report rules anyways.
   -t,--failure-threshold THRESHOLD
                            Exit with failure code only when rules with a
                            severity above THRESHOLD are violated. Accepted
@@ -203,6 +205,7 @@ override:
   info: [string]                        # list of rules
   style: [string]                       # list of rules
 strict-labels: boolean                  # true | false
+disable-ignore-pragma: boolean          # true | false
 trustedRegistries: string | [string]    # registry or list of registries
 ```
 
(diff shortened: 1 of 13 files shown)


zemanlx:
This sounds like good idea, however I think we should not have it as default as it would change current behaviour and I think it can lead to confusion for many users. 

What about having `—ignore-pragma` so CI admins can enable ignoring them?

m-ildefons:
Hi @zemanlx ,
The idea comes from #747 , it is not mine. I should have highlighted the behavior better, but what you describe is as far as I can see exactly what this PR implements. I hope this screenshot gives you a better understanding:
screengrab hadolint-disable-pragma

The default behavior for Hadolint has not changed at all. But you can have Hadolint inform you about encountered ignore pragmas, you can also have Hadolint ignore these pragmas, or both. Each configuration with an optional setting.

zemanlx:
@m-ildefons Oh, I see. I was reading it on my mobile phone and the wrong way (like disable ignoring of pragmas, not disable the ignore type of pragmas). This looks good to me. Sorry for the confusion.

lorenzo (approved):

lorenzo:
@m-ildefons looks great!

PR #759, opened 2022-01-04
Bring env var for trusted registries in sync with docs (#756)

### LINKED ISSUE #756: Setting HADOLINT_TRUSTED_REGISTRIES has no effect
- [x] This is a bug report

### Expected behavior

Running `HADOLINT_TRUSTED_REGISTRIES="something" hadolint Dockerfile` is equivalent to `hadolint --trusted-registry something Dockerfile`.

### Actual behavior

Setting `HADOLINT_TRUSTED_REGISTRIES` has no effect.

The README mentions that this environment variable can be used in lieu of `--trusted-registry`:

> In addition to config files, Hadolint can be configured with environment variables.
> 
> ```
> HADOLINT_TRUSTED_REGISTRIES              # comma separated list of registry urls
> ```

I glanced at `Environment.hs` and it looks like it's [checking for `HADOLINT_ALLOWED_REGISTRIES`](https://github.com/hadolint/hadolint/blob/d16f342c8e70fcffc7a788d122a1ba602075250d/src/Hadolint/Config/Environment.hs#L30) instead.

I'm happy to open a PR for this, just not sure in which direction it should be fixed. Probably the environment variable and README should be updated?

Hadolint version:

```bash
Haskell Dockerfile Linter v2.8.0-0-g398770f-dirty
```

evhan:
### What I did

Renamed the configuration variable from `HADOLINT_{ALLOWED => TRUSTED}_REGISTRIES` in order to match the documented behaviour.

Closes #756.

### How I did it

Just changed the relevant variable string values.

Didn't update any of the procedure names, since they still seem correct; they are still handling the list of allowed registries.

### How to verify it

Run:

```
HADOLINT_TRUSTED_REGISTRIES=good.example.com hadolint /dev/stdin <<< 'FROM bad.example.org/image:v1.0'
```

Expected output:

```
/dev/stdin:1 DL3026 error: Use only an allowed registry in the FROM image
```

### DIFF
--- a/src/Hadolint/Config/Environment.hs
+++ b/src/Hadolint/Config/Environment.hs
@@ -27,7 +27,7 @@ getConfigFromEnvironment =
     <*> getOverrideList "HADOLINT_OVERRIDE_INFO"
     <*> getOverrideList "HADOLINT_OVERRIDE_STYLE"
     <*> getOverrideList "HADOLINT_IGNORE"
-    <*> getAllowedSet "HADOLINT_ALLOWED_REGISTRIES"
+    <*> getAllowedSet "HADOLINT_TRUSTED_REGISTRIES"
     <*> getLabelSchema "HADOLINT_REQUIRE_LABELS"
     <*> maybeTruthy "HADOLINT_STRICT_LABELS"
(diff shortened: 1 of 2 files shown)


lorenzo:
thanks

Issue #737, opened 2021-11-23
hadolint doesn't validate FROM instruction order in Dockerfile as it should


ye:
- [x] This is a bug report
- [x] This is a feature request
- [x] I searched existing issues before opening this one

### Expected behavior

The `FROM` instruction initializes a new build stage and sets the Base Image for subsequent instructions. As such, a valid Dockerfile must start with a FROM instruction (with the only exceptions of comment lines or `ARG`). 

https://docs.docker.com/engine/reference/builder/#from

However, currently `hadolint` doesn't check for that instruction sequence, which is a breaking failure error for `docker build` if this `FROM` precedence rule isn't followed.

### Actual behavior

halolint should error out on any Dockerfile that doesn't start with comments / ARG lines. It doesn't as of version 2.8.0.

### Steps to reproduce the behavior

<!--
Describe the exact steps to reproduce. If possible, provide a *minimum*
reproduction example; take into account that others do not have access
to your private images, source code, and environment.

REMOVE SENSITIVE DATA BEFORE POSTING (replace those parts with "REDACTED")
-->

**Output of `hadolint --version` or
  `docker run --rm hadolint/hadolint hadolint --version` or
  `docker run --rm ghcr.io/hadolint/hadolint hadolint --version`:**

It looks like the Homebrew installed `hadolint` doesn't properly display a version, another bug for the recipes. 

```bash
$ hadolint --version
Haskell Dockerfile Linter UNKNOWN
$ brew info hadolint
hadolint: stable 2.8.0 (bottled)
Smarter Dockerfile linter to validate best practices
https://github.com/hadolint/hadolint
/opt/homebrew/Cellar/hadolint/2.8.0 (5 files, 82MB) *
  Poured from bottle on 2021-11-22 at 17:24:54
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/hadolint.rb
License: GPL-3.0-only
$ cat <<EOF | hadolint -
LABEL maintainer="ye@example.com"
FROM python:3.9-slim-bullseye

ARG DEBIAN_FRONTEND=noninteractive
EOF
$ echo $?
0
cat <<EOF | docker build -f - .
LABEL maintainer="ye@example.com"
FROM python:3.9-slim-bullseye

ARG DEBIAN_FRONTEND=noninteractive
EOF
[+] Building 0.5s (2/2) FINISHED
 => [internal] load build definition from Dockerfile                                                                                                                                                                                     0.3s
[...]

PR #763, opened 2022-01-11
DL3061: Check for instruction order

m-ildefons:
Not all dockerfiles are valid just because all instructions in the
dockerfile are individually valid. For example, a `RUN` instruction may
not precede the first `FROM` instruction and so on. The new rule DL3061
checks that the order of the instructions is not completely bogus. That
means the Dockefile must begin with a comment, an `ARG` or a `FROM`
instruction.

- Add new rule DL3061 to check for instruction order
- Add tests for rule DL3061
- Fix regression tests wrt. rule DL3061

fixes: #737

### How to verify it
The new rule should alert in case the beginning of a dockerfile messes up instruction order such that a build can not succeed, e.g.:
```Dockerfile
RUN foo bar  # this is bad
FROM alpine:3
RUN bar foo
```
But should of course not alert in case a Dockerfile is valid and can be build:
```Dockerfile
# comments are ok
ARG BUILD_ARGUMENTS_OK=true
FROM alpine:3  # FROM must precede other instructions
LABEL foo "bar" # this is ok now
RUN something  # other instructions can follow
```

Thanks to @ye for suggesting this rule.


aaronfi-procore:
Hi @lorenzo this rule is too restrictive, it needs to allow for comments.
i.e. it ended up breaking our production build because our Dockerfile is:

```
# syntax = docker/dockerfile:1.0-experimental

FROM node:16-alpine3.13 AS builder
WORKDIR /app
COPY ./package*.json ./

# etc, etc
```

We need the first comment to remain in place because it is a pragma.  Without it the docker image build itself will fail.

Issue #630, opened 2021-05-05
Allow for Wildcards in the trustedRegistries Config


nxf5025:
- [ ] This is a bug report
- [x] This is a feature request
- [ ] I searched existing issues before opening this one

### Expected behavior
Allow for wildcards in trustedRegistries config. In our use case we are constantly creating new Artifactory repositories which resolve as [repo-name].artifactory-url.com. The ability to use wildcards would allow us to enforce all images are sourced from Artifactory without having to update the config each time a new repo is created.

```
trustedRegistries:
  - *.artifactory-url.com
```

This comment maps out a similar idea, but for a different use case: https://github.com/hadolint/hadolint/issues/309#issuecomment-480827678

lorenzo:
are there use cases for wildcards other than subdomains? I would prefer not to implement a regex-based solution for a narrower case.

nxf5025:
In my situation I can't think of any other use cases unless you wanted to bulk change the severity with something like:
```
override:
  error:
    - DL*
  warning:
    - SC*
  info:
...
```
However I can't imagine this being be too common of a use case.

19D:
i have a use case for this feature. my project hosts several applications with their own dockerfiles. i would like to specify that in order to deploy to production, regardless of how many images we have available, you need to use an image matching a pattern like `one.trusted.registry.com/approvedDir/approvedImages/*`

lorenzo:
Would matching in the prefix be enough?

19D:
unfortunately not. our usage requires a deeper path to specify images.

PR #767, opened 2022-01-22
DL3026: Trusted registries wildcard domains

cronik:
### What I did

Add support for trusted registry domain wildcards. Private registries like
Artifactory use sub-domain for Docker registries so this will allow users to
allow any repository under a single domain (`--trusted-registry "*.megacorp.com"`).

Refs #630

### How I did it

Modified DL3026 isAllowed to search the registries set for entries that are either prefixed or suffixed with "*". For those
entries the item is matched as a suffix or prefix respectively. 

This is my very first attempt at Haskell, so please let me know if there is a better way to go about it.

### How to verify it

Tests have been added to `DL3026Spec` tho validate the matching rules.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,6 +231,7 @@ ignored:
 trustedRegistries:
   - docker.io
   - my-company.com:5000
+  - "*.gcr.io"
 ```
 
 If you want to override the severity of specific rules, you can do that too:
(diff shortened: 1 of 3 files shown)


lorenzo:
thanks!!

PR #770, opened 2022-02-05
Switch to non-rate limited ghcr.io hosted container for pre-commit

dbast:
### What I did

Did run hadolint-docker via pre-commit.

### How I did it

By using a public Github Action runner, which failed due to dockerhub rate limits.

### Solution

Updating the .pre-commit-hooks.yaml to use the ghcr.io hosted public hadolint image, which is not rate limited.


dbast:
ping

lorenzo:
thanks!

PR #777, opened 2022-02-23
Remove duplicate extension

Lunkentuss:
### What I did
Removed the second redundant `PatternSynonyms` default extension in package.yaml.


lorenzo:
thanks!

PR #778, opened 2022-02-27
config: respect NO_COLOR specification

m-ildefons:
- Decide whether to print ANSI color codes or not based on the
  _presence_, not the value of the NO_COLOR environment variable

The specification provided by https://no-color.org regarding the
interpretation of the NO_COLOR environment variable specifically state
that the value does not matter, only whether it is set or not should
change program behavior.
Therefore, Hadolint should respect that and ignore the value.

### How to verify it
Assuming no other settings are in place, `NO_COLOR= hadolint Dockerfile` should result in uncolored output just like `NO_COLOR=true hadolint Dockerfile` and `NO_COLOR=n hadolint Dockerfile`. However `unset NO_COLOR; hadolint Dockerfile` should give colored output.

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,7 +281,7 @@ docker run --rm -i -v /your/path/to/hadolint.yaml:/.config/hadolint.yaml ghcr.io
 In addition to config files, Hadolint can be configured with environment
 variables.
 ```bash
-NO_COLOR=1                               # Truthy value e.g. 1, true or yes
+NO_COLOR=1                               # Set or unset. See https://no-color.org
 HADOLINT_NOFAIL=1                        # Truthy value e.g. 1, true or yes
 HADOLINT_VERBOSE=1                       # Truthy value e.g. 1, true or yes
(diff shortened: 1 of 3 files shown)


lorenzo:
the test fails on windows for some weird reason

m-ildefons:
Hi,
yeah I've seen it, it's not weird, but I'm not sure how to resolve it. The problem is as follows:
no-color.org specifies that the value of `$NO_COLOR` is irrelevant and only whether it is set or not counts. However on unixoids an environment variable can be set to the empty string and count as set, but on Windows setting an environment variable to the empty string is the same as it being unset. Therefore on Windows, the expected outcome of setting `$NO_COLOR` to the empty string is color in the output whereas on unixoids there should be no color in the output in that case.

Maybe I can query what platform Hadolint is running on and adjust the tests behavior accordingly? The case where the setting is applied because the environment variable is set to the empty string only needs to be tested on unixoids because on Windows that is "impossible :copyright: :tm:".

lorenzo:
@m-ildefons yeah, I think detecting the platform during the test and skipping that one on windows is best

lorenzo:
great, thanks again!

Issue #776, opened 2022-02-23
Update shellcheck to 0.8


bendem:
- [ ] This is a bug report
- [x] This is a feature request
- [x] I searched existing issues before opening this one

### Expected behavior

Using `$'\n'` in `RUN`, I should be getting SC3003 and be able to ignore it.

### Actual behavior

I'm getting old shellcheck warnings like SC2039 which have been split into multiple ones. 

### Steps to reproduce the behavior

```bash
> docker run --rm --pull=always -i hadolint/hadolint hadolint - <<EOF
FROM scratch
RUN echo $'\n'
EOF

latest: Pulling from hadolint/hadolint
Digest: sha256:1b47f3192d2c69c622863b84d5cccbfa0c1fac9022bfc297aa931e6d44e51608
Status: Image is up to date for hadolint/hadolint:latest
-:2 SC2039 warning: In POSIX sh, $'..' is undefined.
```

**Additional environment details (OS, stack version, etc.)**

N/A

PR #780, opened 2022-03-01
dependencies: bump shellcheck and language docker

m-ildefons:
- bump language-docker to 10.4.1
- bump shellcheck to 0.8.0

This should bring the newest features of Shellcheck and the latest fixes in the Dockerfile parser to Hadolint as well. May break backwards compatibility because some of Shellcheck's warning numbers (SCxxxx) have changed.

fixes: #776


lorenzo:
thanks!

PR #794, opened 2022-03-16
Experimenting with docker ARM builds

lorenzo:


lorenzo:
I'll close this as it is a dead-end for github actions

Issue #795, opened 2022-03-16
Validation fail with DL3061 error when Dockerfile begin with comment that start with 'syntax'  


StefanoGuerrini:
- [x] This is a bug report
- [ ] This is a feature request
- [x] I searched existing issues before opening this one

### Expected behaviour
Success

### Actual behaviour
DL3061 error: Invalid instruction order. Dockerfile must begin with `FROM`, `ARG` or comment.

### Steps to reproduce the behaviour
A Dockerfile that begin with a comment that starts with 'syntax'

Run:
`docker run --rm -i hadolint/hadolint:latest hadolint "$@" - < Dockerfile`

Version:
  `Haskell Dockerfile Linter v2.9.1-0-g42d0378-dirty`

Error:
```bash
-:1 DL3061 error: Invalid instruction order. Dockerfile must begin with `FROM`, `ARG` or comment.
```

**Dockerfile** 
```Dockerfile
# syntax=something
ARG FOO=BAR
FROM debian:8
```

**Additional environment details (OS, stack version, etc.)**
Using `hadolint/hadolint:2.8.0-debian` works

zeeveener:
This affects all the `latest-X` and `v2.9.1-X` versions. The solution, as mentioned in the description, is to pin the version to `v2.8.0-X`

PR #797, opened 2022-03-17
DL3061: fix false positive on pragmas

m-ildefons:
- Fix false positive error DL3061 when encountering a pragma before a
  `FROM` instruction.

In particular `# syntax = docker/dockerfile...` but also `# shell = ...`
pragmas can appear before the first `FROM` instruction without error.

fixes: #795

### How to verify it
This should no longer produce false positive error DL3061 messages:
```Dockerfile
# syntax = docker/dockerfile:1.0-experimental

FROM node:16-alpine3.13 AS builder
WORKDIR /app
COPY ./package*.json ./
```
Thanks to @aaronfi-procore and @subodhdharma for reporting this issue.


lorenzo:
thanks!

Issue #798, opened 2022-03-18
3061: False positive for ONBUILD commands


cronik:
- [x] This is a bug report
- [ ] This is a feature request
- [ ] I searched existing issues before opening this one

### Expected behavior

3061 should not be triggered for ONBUILD commands.

### Actual behavior

Hadolint rule 3061 is triggered for valid ONBUILD commands.

### Steps to reproduce the behavior

```shell
❯ docker run --rm -i hadolint/hadolint:2.9.1-alpine < Dockerfile
-:2 DL3061 error: Invalid instruction order. Dockerfile must begin with `FROM`, `ARG` or comment.
```

**Output of `hadolint --version` or
  `docker run --rm hadolint/hadolint hadolint --version` or
  `docker run --rm ghcr.io/hadolint/hadolint hadolint --version`:**

```bash
❯ docker run --rm -i hadolint/hadolint:2.9.1-alpine hadolint --version
Haskell Dockerfile Linter v2.9.1-0-g42d0378-dirty
```

**Dockerfile** (if relevant)

```Dockerfile
FROM busybox:1.35
ONBUILD COPY foo.txt /
```

**Additional environment details (OS, stack version, etc.)**

PR #799, opened 2022-03-19
processing: apply onbuild rule wrapper per rule

m-ildefons:
- Apply `onbuild` wrapper per-rule rather than on all rules combined
- Apply `onbuild` wrapper only for rules needing it
- Inline rules _and_ their `onbuild` wrapped counterparts
- Test for DL3047 also when inside an `ONBUILD` statement

The `onbuild` wrapper makes sure a rule is also applied to the AST
branch inside an `ONBUILD` instruction. But this is not necessary for
all rules, for example rules dealing exclusively with `FROM`
instructions need not be applied because that would be invalid syntax
anyways.
Moving the `onbuild` wrapper to apply to each rule individually instead
of the fold of all rules permits granular control over which rules are
applied inside of an `ONBUILD` instruction and crucially also enables
the ability to not apply rules on the AST branch inside of an `ONBUILD`
instruction, if so desired.
In addition to that the inlining for the rules has been adjusted such
that all rules and their `onbuild` wrapped counterparts are inlined into
the fold, before that the fold between the unwrapped and wrapped rules
were both not inlined themselves.

fixes: #798

### DIFF
--- a/src/Hadolint/Process.hs
+++ b/src/Hadolint/Process.hs
@@ -103,22 +103,10 @@ analyze ::
 analyze config =
   AnalisisResult
     <$> Hadolint.Pragma.ignored
-    <*> Foldl.premap parseShell (failures config <> onBuildFailures config)
+    <*> Foldl.premap parseShell (failures config)
 
 parseShell :: InstructionPos Text.Text -> InstructionPos Shell.ParsedShell
 parseShell = fmap Shell.parseShell
-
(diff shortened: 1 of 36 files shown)


lorenzo:
thanks, I really like this solution!

Would you mind solving or silencing the hlint warnings?

lorenzo:
I'll take care of the warnings

Issue #766, opened 2022-01-17
TTY output should include another colon


biffen:
- [ ] This is a bug report
- [X] This is a feature request
- [X] I searched existing issues before opening this one

### Expected behavior

This is such a minor thing but it would make such a nice change.

When using Hadolint from Emacs (with `tty` output) I expect Hadolint to follow the file-colon-line-colon convention (I don’t know that there is a name for this convention, but GNU describes it [here](https://www.gnu.org/prep/standards/html_node/Errors.html)) so that Emacs, and probably other tools (IIRC the Atom editor has something similar), can identify the errors. That way each error becomes a link of sorts to the file and line where the error occurred.

(And yes, I know I can configure Emacs to match more patterns, but there’s really nothing specific to match, so I’d rather see this changed in Hadolint.)

### Actual behavior

The last colon is missing and Emacs doesn’t see the errors as errors.

### Steps to reproduce the behavior

```sh
$ hadolint - <<EOF
FROM alpine
EOF
-:1 DL3006 warning: Always tag the version of an image explicitly
```

If the above contained `-:1:` I’d be happier.

**Output of `hadolint --version` or
  `docker run --rm hadolint/hadolint hadolint --version` or
  `docker run --rm ghcr.io/hadolint/hadolint hadolint --version`:**

```bash
Haskell Dockerfile Linter UNKNOWN
```

**Dockerfile** (if relevant)

Not relevant (I think, let me know otherwise).

**Additional environment details (OS, stack version, etc.)**

Not relevant (again, let me know if I’m wrong)

---

BTW, thanks for an excellent tool!

biffen:
Awesome! Thanks!

PR #801, opened 2022-03-20
formatters: Add GNU style formatter

m-ildefons:
Add a GNU style formatter to Hadolint.
This formatter enhances Hadolints ability to produce output for
consumption by other programs, notably the integration with Emacs will
become easier. GNU style error messages are
described here: https://www.gnu.org/prep/standards/html_node/Errors.html

This formatter differs in three ways from the default TTY output format:
  - the GNU style has an additional colon after the line number
  - the GNU style has no color output
  - the GNU style includes the name of the program (`hadolint`) in the
    output, since Hadolint is not an interactive program

Adding a new formatter instead of modifying the existing TTY
formatter avoids breaking backwards compatibility. Since the differences
only matters when consuming Hadolint's output with another program,
configuring Hadolint to use this formatter does not hurt usability
either.

fixes: #766

hadolint-gnu-formatter

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ Available options:
   -V,--verbose             Enables verbose logging of hadolint's output to
                            stderr
   -f,--format ARG          The output format for the results [tty | json |
-                           checkstyle | codeclimate | gitlab_codeclimate |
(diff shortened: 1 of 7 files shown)


lorenzo:
thanks, this is great!

Issue #782, opened 2022-03-03
Using a lower bound for version pinning with alpine apk triggers version pinning warnings and stdout redirection error


stonecharioteer:
- [x] This is a bug report
- [ ] This is a feature request
- [x] I searched existing issues before opening this one

### Expected behavior

This should not have had any errors

### Actual behavior
Using `>=` in version pinning triggers 2 events: hadolint thinks that the version isn't pinned since it uses `>=` instead of `=`, and it also thinks that the STDOUT is redirected twice.

### Steps to reproduce the behavior

I'm using the following in an alpine image.

```Dockerfile
FROM rust:1.59-alpine3.14 AS builder
WORKDIR /app
RUN apk add --no-cache musl-dev>=1.1 pkgconfig>=0.23 openssl>=1.1 libressl-dev>=3.5
```

hadolint version is v2.8.0.

```bash
Dockerfile:3 DL3018 warning: Pin versions in apk add. Instead of `apk add <package>` use `apk add <package>=<version>`
Dockerfile:3 SC2261 error: Multiple redirections compete for stdout. Use cat, tee, or pass filenames instead.
```

lorenzo:
Would you like to contribute a fix for this? It should be simple to do and I can guide you

stonecharioteer:
@lorenzo Yes! I'd love to. How do I get started?

stonecharioteer:
While the [official documentation doesn't specifically show `>=`](https://wiki.alpinelinux.org/wiki/Package_management#Holding_a_specific_package_back), here's a screenshot of me trying it out in an alpine container. 

![2022-03-31-151921_772x407_scrot](https://user-images.githubusercontent.com/11478411/161028172-4d2a9ca7-b286-4972-a3f2-a74282c72742.png)

lorenzo:
@[stonecharioteer](https://github.com/stonecharioteer) Thanks, let's focus first on the first warning. You will need to do this:

* You need to edit this file https://github.com/hadolint/hadolint/blob/master/src/Hadolint/Rule/DL3018.hs#L32
* Repeat the same line I linked to, so that it contains `>=` instead of `=`
* Add a test, like this at the bottom of the file https://github.com/hadolint/hadolint/blob/master/test/Hadolint/Rule/DL3018Spec.hs#L28-L38



Thanks for your help

m-ildefons:
Fixed in v2.15.0

PR #812, opened 2022-04-01
allow lower bound version pinning for apk (DL3018)

stonecharioteer:
This PR addresses #782.

@lorenzo this is still a WIP, since my test is failing. I've never written haskell code before, so perhaps I'm missing something.

### What I did

I've added an or section to the check for DL3018 so that it allows `>=` in `apk add` steps. 

### How I did it

I followed @lorenzo's instructions on #782.

### How to verify it

`stack test`

The tests still fail, I'm not sure how to fix that. I could use some help there. 

Aside: Additionally, this is a thing that could be used in all places where lower bounds are respected by the OS. This is supported even by yum, apt and pip. Should I implement it for those as well?

Thank you for guiding me in fixing this.

### DIFF
--- a/src/Hadolint/Rule/DL3018.hs
+++ b/src/Hadolint/Rule/DL3018.hs
@@ -29,7 +29,7 @@ dl3018 = simpleRule code severity message check
         )
         args
     check _ = True
-    versionFixed package = "=" `Text.isInfixOf` package
+    versionFixed package = "=" `Text.isInfixOf` package || ">=" `Text.isInfixOf` package
     packageFile package = ".apk" `Text.isSuffixOf` package
 {-# INLINEABLE dl3018 #-}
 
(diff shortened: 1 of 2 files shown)


stonecharioteer:
@lorenzo here's the error I get.
2022-04-01-102105_1079x150_scrot

lorenzo:
I'll take a look today

stonecharioteer:
@lorenzo did you get time to take a look?

m-ildefons:
Hi @stonecharioteer ,
The `apk add` command here is first parsed as shell, then parsed by the rule. Since `>` indicates redirection in POSIX shells, you need to make sure your version bounds are either quoted or escaped properly, e.g.:
```Dockerfile
RUN apk add --no-cache \
    "flex>=2.6.4-r1"
```
or in the test suite:
```Haskell
    let dockerfile =
        [ "RUN apk add --no-cache \\",
          "\"flex>=2.6.4-r1\" \\",
          "\"libffi>=3.2.1-r3\""
        ]
```

m-ildefons:
I'm going to close this PR, since it's now in conflict and it seems unlikely that it's going to be fixed anytime soon.
There is a followup PR implementing this fix here: https://github.com/hadolint/hadolint/pull/1201

PR #814, opened 2022-04-06
unpin docker version in pre-commit hook

foolioo:
### What I did
eliminating inconsistency. e.g. for version 2.10.0, pre-commit hooks points to outdated version https://github.com/hadolint/hadolint/blob/47e3022493cb0d51e5af21ee346d4b25bc02bfcd/.pre-commit-hooks.yaml#L7

### How I did it
unpinning the version from the docker hook, behaves in the same manner as the native hook.
this is not the best approach, since each version will pull the latest docker image but will also run on the latest version that is installed on the operating system. the best way to handle the issue is to pin the exact version with each release.

### How to verify it
 set `.pre-commit-config.yaml` as follows
```yaml
  - repo: https://github.com/hadolint/hadolint
    rev: 23b2b76fe7b757658019ca4e080efa208784621b
    hooks:
      - id: hadolint-docker
```
then execute pre-commit and see whether the latest docker images has been pulled
```shell
$ pre-commit run -a hadolint-docker
$ docker run --rm ghcr.io/hadolint/hadolint /bin/hadolint --version
Haskell Dockerfile Linter 2.10.0
```

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -4,7 +4,7 @@
   description: Runs hadolint Docker image to lint Dockerfiles
   language: docker_image
   types: ["dockerfile"]
-  entry: ghcr.io/hadolint/hadolint:v2.9.3 hadolint
+  entry: ghcr.io/hadolint/hadolint hadolint
 - id: hadolint
   name: Lint Dockerfiles
   description: Runs hadolint to lint Dockerfiles


lorenzo:
maybe better to commit the file after each release in a github action?

foolioo:
@lorenzo i think it will not solve the issue. the `pre-commit-config.yaml` must pin the same version as the git tag because pre-commit checkout the repository at the `rev` and then reads the pre-commit-hook.yaml (which is where the version of the docker image is specified)

foolioo:
@lorenzo any other thoughts?

lorenzo:
I’m not familiar with pre-commit hooks. What do you mean by “it checkouts the commit at the rev”? Do you have to keep updating the version locally?

foolioo:
> What do you mean by “it checkouts the commit at the rev”?

from [pre-commit docs](https://pre-commit.com/#pre-commit-configyaml---repos)
> `rev`: the revision or tag to clone at.
>
> A sample repository:
> ```yaml
> repos:
> -   repo: https://github.com/pre-commit/pre-commit-hooks
>     rev: v1.2.3
>     hooks:
>     -   ...
> ```

foolioo:
@lorenzo did i answer your question?

lorenzo:
@foolioo Thanks, I understand what you mean. If this is merged, how do you usually tell docker to update to the latest tag when it is released? Once an image is downloaded to the cache, it will not change anymore

foolioo:
@lorenzo there are 3 ways:
1. you pin and bump the version in the hook with each release (of curse, this can be automated)
2. use pre-commit with the docker image directly, watch for changes, and update accordingly
3. expect the dependency in hadolint to be met

alternatives are enumerated by order of best practice.

lorenzo:
Ok, I understand. I think it makes sense to change this. Thanks for your patience

PR #824, opened 2022-04-22
Update pre-commit hook to use latest docker image

noorul:


lorenzo:
went with a different solution. thanks!

Issue #775, opened 2022-02-22
.hadolint.yaml config is ignored


jornfranke:
- [x] This is a bug report
- [ ] This is a feature request
- [X] I searched existing issues before opening this one

### Expected behavior

hadolint reads the file .hadoolint.yaml and applies the options. E.g. in this case it should especially consider output format "format" in it.

### Actual behavior

hadolint ignores at least the option "format" in the yaml, but most likely all options. If I provide a non-existing file then it complains that it cannot read the file, so I assume it finds the file and content.
If I provide the format via the -f flag then it is correctly produced.

### Steps to reproduce the behavior

```
hadolint -c ./hadolint.yaml Dockerfile
```

Use the following file ./hadolint.yaml
```
failure-threshold: "ignore"
format: "gitlab_codeclimate"
no-color: false
no-fail: true
strict-labels: true
disable-ignore-pragma: false
trustedRegistries:
    - "internal"
    - "internal2"
  ```

It will not output the issues in "gitlab_codeclimate" format, but the default tty.

hadolint version: 2.8.0 
OS: Linux
Gitlab Runner: 14.7.0

**Comment by codezninja** [comment]
Seems like this broke in 2.8.0. Using version 2.7.0 seems to read the config properly. When I used the `-V` option I can see that hadolint reads the config file but it doesn't load any of the values

**Comment by m-ildefons** [comment]
Hi,
thanks for reporting this bug. I believe this is actually the issue described [here](https://github.com/hadolint/hadolint/pull/734) and fixed [here](https://github.com/hadolint/hadolint/commit/2fc50c13b8e23fd0581c3a5a73bc0a51ce1da583).

**Comment by jornfranke** [comment]
It is not only trustedregistries - it seems everything in the config is ignored

**Comment by codezninja** [comment]
@jornfranke it's actually only a few things from what I can tell. `failure-threshold` worked for me in the above example.

**Comment by huornlmj** [comment]
Also reporting failure-threshold not working correctly.  The following within .hadolint.yaml...

```
failure-threshold:
        - info
```

yields...

```
[...]

PR #829, opened 2022-04-29
bugfix: config file key for output format

m-ildefons:
Make Hadolint read the desired output format from the key `format` in
the config file, instead of `output-format`. This is more convenient,
consistent with the command line options and consistent with the
documentation.

fixes: #775


lorenzo:
thanks

PR #839, opened 2022-05-28
bugfix: typo in DL3026Spec.hs

cwimmer:
### What I did
Fixed a typo in DL3026Spec.hs

### How I did it
Swapped two characters

### How to verify it
Read DL3026Spec.hs


lorenzo:
thanks!

PR #845, opened 2022-06-15
DL3013: Fix false positives for VCS, http and local path packages (#389)

### LINKED ISSUE #389: pip install directory detected as DL3013
- [x] This is a bug report

### Expected behavior
This shouldn't be DL3013 as it installs a setup.py in a directory
```
RUN pip install /app
```

### Actual behavior
```
/dev/stdin:1 DL3013 Pin versions in pip. Instead of `pip install <package>` use `pip install <package>==<version>`
```

```
echo "RUN pip install /app"  | hadolint -
```

```bash
Haskell Dockerfile Linter v1.17.5-0-g443423c
```

### LINKED ISSUE #398: Suggest apk add --upgrade when apk upgrade is used
- [ ] This is a bug report

### Suggestion

Suggest to use ``apk add --upgrade`` instead of ``apk upgrade`` to only upgrade a specific package and not all.

See https://github.com/docker-library/dockerfile-validator/blob/master/dockerfile-validator#L156 and https://wiki.alpinelinux.org/wiki/Upgrading_Alpine

Lunkentuss:
### What I did
- Fix false positive when installing package with other supported VCS
  such as: `pip install bzr+ssh://..@1.0.0`, as described in
  https://pip.pypa.io/en/stable/topics/vcs-support/.
- Fix false positive when installing package from local directory:
  `pip install my/install/path`
- Fix false positive when installing package from http/https url:
  `pip install https://foo.bar/baz-1.0.0.zip`

fixes #389

### How I did it
Extended the previous check for a `git+https` vcs scheme to allow any of the supported vcs sources, such as `git+ssh` and `bzr+ssh`, documented in https://pip.pypa.io/en/stable/topics/vcs-support/. Also added passing check if a package source included a forward slash (/) but is not a valid vcs source.

### How to verify it
Compare the supported VCS in this PR from the VCS supported from the pip documentation https://pip.pypa.io/en/stable/topics/vcs-support/. Also, the following Dockerfile illustrates cases where hadolint now runs without any warnings.

```Dockerfile
FROM python:3.10

RUN pip install --no-cache-dir \
  git+ssh://hello@1.3 \
  hello/bar/baz \
  https://foo.bar/baz-1.0.0.zip
```

### DIFF
--- a/src/Hadolint/Rule/DL3013.hs
+++ b/src/Hadolint/Rule/DL3013.hs
@@ -44,12 +44,18 @@ dl3013 = simpleRule code severity message check
         || ["."] `isInfixOf` Shell.getArgs cmd
 
     hasBuildConstraint cmd = Shell.hasFlag "constraint" cmd || Shell.hasFlag "c" cmd
-    versionFixed package = hasVersionSymbol package || isVersionedGit package || isLocalPackage package
-    isVersionedGit package = "git+http" `Text.isInfixOf` package && "@" `Text.isInfixOf` package
+    versionFixed package = hasVersionSymbol package
+      || isVersionedVcs package
+      || isLocalPackage package
+      || isNoVcsPathSource package
+    isVersionedVcs package = isVcs package
+      && "@" `Text.isInfixOf` package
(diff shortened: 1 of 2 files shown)


lorenzo:
thanks!

PR #850, opened 2022-06-26
Add SC3046 in hadolint JSON Schema

nvuillam:
This avoid an error in IDE when updating .hadolint.yml config file


lorenzo:
thanks!

PR #860, opened 2022-07-23
docs: update link to Haskell

chapeupreto:
The Haskell Platform is deprecated since 2022 and is no longer the recommended way of installing Haskell. For the latest recommended way to install Haskell please see the Downloads page.


lorenzo:
Thanks

PR #871, opened 2022-09-07
feature: global ignore pragma

m-ildefons:
Add global variant of the ignore pragma. This allows users to disable rules for the whole dockerfile by using a pragma like:

```dockerfile
# hadolint global ignore=DL3003,SC2035
```

### What I did
- add file-global variant of the ignore pragma
- clean up the pragma parsing code

### How I did it
The file global ignore pragma works almost like the inline ignore pragma. The difference is that it parses into just a set of rules to ignore instead of a map from line-numbers to sets of rules to ignore. Consequently there is no check for line numbers when deciding if a check-failure should be ignored or propagated, all that matters is if that failures code is in that global set of rule codes to ignore.

### How to verify it
The following Dockerfile should not trigger any of the globally ignored rules, but will still trigger all other rules just normal:

```Dockerfile
# hadolint global ignore=DL3002
FROM alpine
USER root
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -333,6 +333,19 @@ RUN cd /tmp && echo "hello!"
 
 The comment "inline ignores" applies only to the statement following it.
 
+## Global ignores
+
+Rules can also be ignored on a per-file basis using the global ignore pragma.
+It works just like inline ignores, excep that it applies to the whole file
+instead of just the next line.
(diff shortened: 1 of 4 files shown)


lorenzo:
Thanks!

PR #900, opened 2022-11-12
fix: false positive in `DL3022`

swarnimarun:
### What I did

Simple patch to fix bug with hadolint raising false positives if the stages are mentioned by index rather than names.
Such as,
```dockerfile
FROM X as stage1
RUN # do some work, likely, build something to copy

FROM Y as stage2
COPY --from=0 bar .
```

### How I did it

I use decimal conversion over string to get the index and compare it to the count of the stages in source, not sure if there was a better way to write the code I just wrote it because I needed it to be fixed.

### How to verify it

Added a couple tests to `DL3022Spec.hs`

#### Note: Let me know if there are any mistakes I will fix em. Though as it's a weekend I might not always be available.

### DIFF
--- a/src/Hadolint/Rule/DL3022.hs
+++ b/src/Hadolint/Rule/DL3022.hs
@@ -3,7 +3,6 @@ module Hadolint.Rule.DL3022 (rule) where
 import qualified Data.Set as Set
 import qualified Data.Text as Text
 import qualified Data.Text.Read as Read
-
 import Hadolint.Rule
 import Language.Docker.Syntax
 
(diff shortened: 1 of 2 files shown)


swarnimarun:
sorry had to add another test as I remembered this case was broken, also fixed it :P

fixed it with a bit of hack (names can't be plain numbered indexes, so it should be perfectly fine, but it's a hack nonetheless), would have had to add more stuff to state otherwise.

lorenzo:
I would change the state to be able to store numbers as well. It feels messy having to parse a number each time you want to check the state.

Do you think you can make that change?

swarnimarun:
@lorenzo let me know if it's fine now, not that good at Haskell. But tried to follow project conventions.

I couldn't come up with anything more beautiful so feel free to make edits. :P

swarnimarun:
Also sorry if I don't follow up immediately a bit loaded on the work side of life.

lorenzo:
looks great, thanks!

Issue #876, opened 2022-09-22
unexpected '$' expecting '`', a new line followed by the next instruction, or the variable name


mloskot:
- [x] This is a bug report
- [ ] This is a feature request
- [x] I searched existing issues before opening this one

### Expected behavior

I'm not certain where is the problem but `docker` is happy to build images based on Windows ServerCore from my `Dockerfile`,
so I'd expect the linter accepts the `EXPOSE` port range specification in it (`Dockerfile` below):

```
${APP_PORT}-${APP_PORT_MAX}/tcp
```

### Actual behavior

```
bash-5.1# hadolint --no-fail --no-color test/Dockerfile 
test/Dockerfile:23:13 unexpected '$' expecting '`', a new line followed by the next instruction, or the variable name
```

I understand the line `${APP_PORT}-${APP_PORT_MAX}/tcp` in this bit

```
EXPOSE `
${APP_PORT}/tcp `
${APP_PORT}-${APP_PORT_MAX}/tcp
```

can be a bit unusual specification of port range, but Docker is fine about.

Simplifying it this way does not help:

```
EXPOSE ${APP_PORT}/tcp ${APP_PORT}-${APP_PORT_MAX}/tcp
```

### Steps to reproduce the behavior

**Output of `hadolint --version` or
  `docker run --rm hadolint/hadolint hadolint --version` or
  `docker run --rm ghcr.io/hadolint/hadolint hadolint --version`:**

```bash
# hadolint --version
Haskell Dockerfile Linter 2.10.0
```

**Dockerfile** (if relevant)

```Dockerfile
# escape=`
ARG BASE_IMAGE
ARG BASE_IMAGE_TAG

FROM ${BASE_IMAGE}:${BASE_IMAGE_TAG}
ARG ENTRYPOINT
ARG APP_VERSION
ENV APP_VERSION=${APP_VERSION}
WORKDIR /
COPY `
${ENTRYPOINT} `
scripts\app.ps1 `
./

ENV `
APP_PATH=C:\app.ps1 `
APP_PORT=4330 `
APP_PORT_MAX=4340

EXPOSE `
${APP_PORT}/tcp `
${APP_PORT}-${APP_PORT_MAX}/tcp

VOLUME ["D:", "Z:"]

ENTRYPOINT ["powershell.exe", "-NoProfile", "-ExecutionPolicy", "Bypass", "C:\\appps1;"]
```

m-ildefons:
Hi Mateusz,
[...]

PR #902, opened 2022-11-15
DL3011: fixup changes in AST

m-ildefons:
Fix up usage of the AST in DL3011 corresponding to the changes to the AST with respect to port ranges in `EXPOSE` instructions.

requires: https://github.com/hadolint/language-docker/pull/85
fixes: https://github.com/hadolint/hadolint/issues/876


lorenzo:
great, thanks!

Issue #899, opened 2022-11-10
Not able to install from binary


andrybicio:
- [x] This is a bug report

My OS:
- Ubuntu 20.04LTS in WSL2 (Windows 10)

```
$  stack ghc -- --version
The Glorious Glasgow Haskell Compilation System, version 9.0.2
```

I wanted to compile and install hadolint on my machine, via the instructions provided in the README. 
```
git clone https://github.com/hadolint/hadolint \
  && cd hadolint \
  && stack install
```
However, the last command (i.e., `stack install`) returns:
```
   Error parsing targets: The specified targets matched no packages.
   Perhaps you need to run 'stack init'?
```

Doing `stack init` as suggested, leads to:
```
Looking for .cabal or package.yaml files to use to init the project.
Using cabal packages:
- ./

(OMITTED)

* Partially matches https://raw.githubusercontent.com/commercialhaskell/stackage-snapshots/master/lts/3/22.yaml
    HsYAML not found
        - hadolint requires >=0
    ShellCheck not found
        - hadolint requires >=0.8.0
    colourista not found
        - hadolint requires >=0
    deepseq version 1.4.1.1 found
        - hadolint requires >=1.4.4 && <1.5
    directory version 1.2.2.0 found
        - hadolint requires >=1.3.0
    gitrev version 1.1.0 found
        - hadolint requires >=1.3.1
    hspec version 2.1.10 found
        - hadolint requires >=2.8.3
    ilist not found
        - hadolint requires >=0
    language-docker not found
        - hadolint requires >=11.0.0 && <12
    megaparsec not found
        - hadolint requires >=9.0.0
    optparse-applicative version 0.11.0.2 found
        - hadolint requires >=0.14.0
    parsec version 3.1.9 found
        - hadolint requires >=3.1.14
    prettyprinter not found
        - hadolint requires >=1.7.0
    Using package flags:
        - hadolint: FlagName "static" = True

Selected resolver: https://raw.githubusercontent.com/commercialhaskell/stackage-snapshots/master/nightly/2022/11/8.yaml
Resolver 'https://raw.githubusercontent.com/commercialhaskell/stackage-snapshots/master/nightly/2022/11/8.yaml' does not have all the packages to match your requirements.
    colourista not found
        - hadolint requires >=0
    Using package flags:
        - hadolint: FlagName "static" = True

This may be resolved by:
    - Using '--omit-packages' to exclude mismatching package(s).
[...]

PR #903, opened 2022-11-15
docs: use cabal instead of stack in readme

m-ildefons:
Use cabal-install instead of stack in the readme, since stack support was removed along with stack.yaml configuration.

fixes: #899
Signed-off-by: Moritz Röhrich <moritz@ildefons.de>


lorenzo:
Thanks so much!
Would you mind pleasing the linter?

lorenzo:
thanks!

Issue #911, opened 2022-12-05
Dont throw DL3014 when using "apt-get --quiet --quiet"


SpikePy:
- [x] This is a bug report
- [ ] This is a feature request
- [ ] I searched existing issues before opening this one

### Expected behavior

when using `apt-get --quiet --quiet` no warning DL3014 should be thrown

### Actual behavior

```
DL3014 warning: Use the `-y` switch to avoid manual input `apt-get -y install <package>`
```

### Steps to reproduce the behavior

hadolint this Dockerfile

```
FROM debian:stable-slim
RUN apt-get update && apt-get install --quiet --quiet sl
```

throws the given error but the documentory aof apt-get says that "--quiet --quiet" implicates --yes so this should not be shown!

PR #913, opened 2022-12-20
DL3014: Fix false positive on double quiet

m-ildefons:
`apt-get` can take the `--quiet` or `-q` argument to increase its quietness level. At the maximum of 2, this will also imply `assume-yes` i.e. the `-y` flag. Therefore if the necessary quietness level has been configured by command line parameters, DL3014 should not alert the user to the missing `-y` flag.

fixes: #911

Signed-off-by: Moritz Röhrich <moritz.rohrich@suse.com>

### How to verify it
Unittests are included. In any case, the following Dockerfile should no longer produce warning DL3014:

```Dockerfile
FROM debian:stable-slim
RUN apt-get update && apt-get install --quiet --quiet sl
```


lorenzo:
thanks!

lorenzo (approved):

PR #940, opened 2023-02-27
fix: misidentification of pipenv (#366)

### LINKED ISSUE #366: `Pin versions in pip` linting error
- [x] This is a bug report

When my Dockerfile looks like this:
```
RUN pipenv run python setup.py install
```
I get this:
```
$ hadolint Dockerfile
Dockerfile:1 DL3013 Pin versions in pip. Instead of `pip install <package>` use `pip install <package>==<version>`
```

I shouldn't get a linter error in this case as there is no `pip install` command in that line.

```bash
echo "RUN pipenv run python setup.py install" > Dockerfile
hadolint Dockerfile
```

kthy:
### What I did

Check that when we detect `pip` it's not actually `pipenv`. This fixes #366.

### How I did it

Add an extra `&& not (…)` check.

### How to verify it

Run Hadolint on this Dockerfile:

```
FROM python:3.10
RUN pip install --no-cache-dir pipenv==2023.2.18 \
 && python3 -m venv /venv
RUN pipenv install --python 3.10 --keep-outdated --ignore-pipfile \
 && rm -rf /root/.cache/pipenv
```

It raises DL3013 on the current version of Hadolint and nothing on my fork.

(Caveat: I had to compile with a newer `base` package, but the change should be simple enough that it makes no difference.)

### DIFF
--- a/src/Hadolint/Shell.hs
+++ b/src/Hadolint/Shell.hs
@@ -241,6 +241,7 @@ isPipInstall cmd@(Command name _ _) = isStdPipInstall || isPythonPipInstall
   where
     isStdPipInstall =
       "pip" `Text.isPrefixOf` name
+        && not ("pipenv" `Text.isPrefixOf` name)
         && ["install"] `isInfixOf` getArgs cmd
     isPythonPipInstall =
       "python" `Text.isPrefixOf` name


m-ildefons:
Hi Kristian,
thanks a lot for this contribution. Would you be so kind and add tests in `test/Hadolint/Rule/DL3013Spec.hs` for this case too?
Other than that the code looks good to me.

kthy:
This is my first time writing Haskell, so not really sure what you expect in the `it "…"` string. I've made a guess. 😊

m-ildefons:
That was quick :smile: , thanks a lot.

`it` is a test item and the string should be something to uniquely identify that particular test within the file. It can be used to re-run individual tests. Together with `describe` it should give an idea of what the test is trying to achieve, so you guessed quite good.
For more advanced usage, see the docs: https://hspec.github.io/writing-specs.html#using-describe-and-it

PR #954, opened 2023-04-26
Improve parsing of gem install flag patterns to reduce false positives

mkusaka:
### What I did
In Hadolint, there are several gem install flag patterns that result in false positives.

Warnings should not be displayed for the following patterns:

```dockerfile
RUN gem install bundler:2.0.1 -- --use-system-libraries true
RUN gem install bundler:2.0.1 --use-system-libraries true
RUN gem install bundler:2.0.1 --use-system-libraries=true
```

In this PR, I have made adjustments so that these patterns can also be parsed correctly.

### How I did it
I modified the processing to remove options and their subsequent arguments from the list of arguments.

### How to verify it
I have added tests related to the additional patterns.


lorenzo:
thanks

PR #1003, opened 2024-02-22
DL3060 should be suppressed if cache mounts are used

slimslenderslacks:
### What I did

```
## this should violate DL3060 because there's no yarn cache clean
RUN yarn install
```

On the other hand, this does not have yarn cache clean but is still okay because the cache will not be included in the final image and removing the cache defeats the purpose of using the cache mount.

```
RUN --mount=type=cache,target=/usr/local/share/.cache/yarn/v6 yarn install
```

### How I did it

I added `RunFlags` to the check.  I checked whether the RunMount contains any CacheMounts.

### How to verify it

The second `RUN` line above should not raise DL3060.


lorenzo:
thanks!

PR #1013, opened 2024-04-24
docs: Update TOC with Ignoring Rules section

tanmay-pnaik:
### What I did

Update TOC with Ignoring Rules section


lorenzo:
thanks again

PR #1015, opened 2024-05-21
Update README.md

afmpjr:
+ Quick fix to correctly run the project using docker
+ script for user experience convenience

### What I did
Tested locally using the current docker instructions and didn't work as expected.

### How I did it
Modified the `docker run` command by adding the `Dockerfile` volume
Created a script for user convenience

### How to verify it
Simply follow the provided instructions

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,9 +53,36 @@ platforms.
 Just pipe your `Dockerfile` to `docker run`:
 
 ```bash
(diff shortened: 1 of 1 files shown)


m-ildefons:
This is a lot more complicated and the existing instructions work just fine.

If you continue to have trouble with the existing instructions, please file an issue describing what exactly isn't working.

PR #344, opened 2019-08-29
Ignore unsatisfiable shellcheck error 1090

ianfixes:
According to https://github.com/koalaman/shellcheck/wiki/SC1090 the fix is to put a shell comment "directive" to assist -- "# shellcheck source=/dev/null".  This can't work in a Dockerfile because `#` is also the Docker comment character.  So with no obvious workaround, the error must simply be ignored completely.

* fixes #343

Issue #343, opened 2019-08-29
SC1090 can't be satisfied -- it wants shellcheck directives (which are shell comments)


ianfixes:
- [x] This is a bug report
- [ ] This is a feature request
- [x] I searched existing issues before opening this one

### Minimal example

Consider this `mini.Dockerfile`

```Dockerfile
FROM alpine:3
SHELL ["/bin/bash", "-c"]
ENV NVM_DIR=/usr/local/nvm
RUN source $NVM_DIR/nvm.sh
```

Linting fails:

```console
$ docker run --rm -i hadolint/hadolint:latest /bin/hadolint - < mini.Dockerfile 
/dev/stdin:4 SC1090 Can't follow non-constant source. Use a directive to specify location.
```

### What does `shellcheck` say?

According to https://github.com/koalaman/shellcheck/wiki/SC1090 the fix is to put a shell comment "directive" to assist:

```shell
# shellcheck source=src/lib.sh
. "$(find_install_dir)/lib.sh"
```

> If you don't care that ShellCheck is unable to account for the file, specify `# shellcheck source=/dev/null`.

### But that doesn't work...

`mini2.Dockerfile`

```Dockerfile
FROM alpine:3
SHELL ["/bin/bash", "-c"]
ENV NVM_DIR=/usr/local/nvm
RUN # shellcheck source=/dev/null
RUN source $NVM_DIR/nvm.sh
```

```console
$ docker run --rm -i hadolint/hadolint:latest /bin/hadolint - < mini2.Dockerfile 
/dev/stdin:4 SC1072 Expected a command. Fix any mentioned problems and try again.
/dev/stdin:5 SC1090 Can't follow non-constant source. Use a directive to specify location.
```

The old "Now you have 2 problems"...

`mini3.Dockerfile`

```Dockerfile
FROM alpine:3
SHELL ["/bin/bash", "-c"]
ENV NVM_DIR=/usr/local/nvm
# shellcheck source=/dev/null
RUN source $NVM_DIR/nvm.sh
```

```console
$ docker run --rm -i hadolint/hadolint:latest /bin/hadolint - < mini3.Dockerfile 
/dev/stdin:5 SC1090 Can't follow non-constant source. Use a directive to specify location.
```

### Expected behavior

Either drop the error entirely or provide a valid way to work around it.

### Version

**Output of `hadolint --version` or
  `docker run --rm hadolint/hadolint hadolint --version`:**

```bash
[...]

PR #1020, opened 2024-05-31
Ignore unsatisfiable shellcheck error 1091

ghoneycutt:
This check is related to SC1090 which is already skipped in PR #344 and background in issue #343.

### What I did

Add SC1091 to list of shellcheck entries to ignore.

### Why I did it

Because SC1090 and SC1091 are essentially the same and SC1090 is already excluded.

### How I did it

Simple addition to existing list.

### How to verify it

Use this `Dockerfile`

```dockerfile
FROM alpine:3
RUN . /etc/sysconfig/somefile
```

before

```sh
$ hadolint Dockerfile
Dockerfile:2 SC1091 info: Not following: File not included in mock.
```

after

```sh
~/.local/bin/hadolint Dockerfile ; echo $?
0
```

### DIFF
--- a/src/Hadolint/Shell.hs
+++ b/src/Hadolint/Shell.hs
@@ -79,7 +79,8 @@ shellcheck (ShellOpts sh env) (ParsedShell txt _ _) =
     script = Text.unpack $ "#!" <> extractShell sh <> "\n" <> printVars <> txt
     exclusions =
       [ 2187, -- exclude the warning about the ash shell not being supported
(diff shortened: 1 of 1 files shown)


markholland:
Came across this today. Tried adding it to .hadolint.yaml ignored list but now VSCode complains that SC1091 isn't in the yaml-schema for hadolint.

markholland (approved):

lorenzo:
thanks

markholland:
When will this be released?

PR #1034, opened 2024-08-28
DL3048 accepting LABEL keys that contain underscore(s) or slash(es)

mschmeitz:
Underscores and slashes should be accepted as part of a LABEL key. See  https://docs.docker.com/engine/manage-resources/labels/#key-format-recommendations

### DIFF
--- a/src/Hadolint/Rule/DL3048.hs
+++ b/src/Hadolint/Rule/DL3048.hs
@@ -36,4 +36,4 @@ hasConsecutiveSeparators :: Text.Text -> Bool
 hasConsecutiveSeparators l = ".." `Text.isInfixOf` l || "--" `Text.isInfixOf` l
 
 validChars :: String
-validChars = ['0'..'9'] ++ ['a' .. 'z'] ++ ['.', '-']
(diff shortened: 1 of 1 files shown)


Bond-009:
From your own link (the guidelines following the main body):
> Label keys should begin and end with a lower-case letter and should only contain lower-case alphanumeric characters, the period character (.), and the hyphen character (-). Consecutive periods or hyphens aren't allowed.

i.e. while underscores (_) and slashes (/) are technically allowed, you shouldn't use them

mschmeitz:
I agree, you should not use them. But it does not say, you may not use them. First sentence in the Docker Docs:  https://docs.docker.com/engine/manage-resources/labels/#key-format-recommendations

_A label key is the left-hand side of the key-value pair. Keys are alphanumeric strings which may contain periods (.), underscores (_), slashes (/), and hyphens (-)._

So, either merge or decline.

Bond-009:
First of all: this is a linter, the intention is helping to build best practice Docker images, not just valid ones.
You yourself agree underscores and slashes shouldn't be used in label keys.
Secondly I'm not a maintainer of this program, so I'm not in a position to accept or close this request. What I can do is give some context why the rule is the way it is and why this change doesn't make sense.

Issue #1028, opened 2024-07-04
Allow text in ignore comments


avazhnov-twilio:
- [ ] This is a bug report
- [x] This is a feature request
- [x] I searched existing issues before opening this one

Currently such syntax is supporting by `hadolint` for [Inline ignores](https://github.com/hadolint/hadolint#inline-ignores):

```sh
# hadolint ignore=DL3003,SC1035
```

When I use Shellcheck, I prefer expanding a comment with text description, which is more human-readable:

```sh
# shellcheck disable=SC1091  # File not included in mock
```

Can we have similar logic for `hadolint`: to ignore whitespaces and custom text (for example, starting with `#`) in the same line after `ignore=ID1,ID2`?

### Expected behavior

Line

```Dockerfile
# hadolint ignore=SC1091  # File not included in mock
```

makes `hadolint` ignore `SC1091` in the next line.

### Actual behavior

Code

```Dockerfile
# hadolint ignore=SC1091  # File not included in mock
RUN python3 -m venv .venv && . .venv/bin/activate && pip3 install --no-cache-dir -r requirements.txt
```

gives a message like: "Dockerfile:7 SC1091 info: Not following: File not included in mock."

### Steps to reproduce the behavior

Create a `Dockerfile` like

```Dockerfile
FROM python:3

RUN groupadd -r myapp && useradd --no-log-init --system --gid myapp --create-home myapp
USER myapp

# hadolint ignore=SC1091  # File not included in mock
RUN python3 -m venv .venv && . .venv/bin/activate && pip3 install --no-cache-dir -r poetry
```

Check it with `hadolint`, it will give you a message: "Dockerfile:7 SC1091 info: Not following: File not included in mock."

**Output of `hadolint --version`:**

```bash
Haskell Dockerfile Linter 2.12.0
```

**Dockerfile** (if relevant)

```Dockerfile
# hadolint ignore=SC1091  # File not included in mock
RUN python3 -m venv .venv && . .venv/bin/activate && pip3 install --no-cache-dir -r requirements.txt
```

**Additional environment details (OS, stack version, etc.)**

Probably my request may also close https://github.com/hadolint/hadolint/issues/997.

Issue #997, opened 2024-01-30
Allow trailing whitespace in ignore comments


PeterStolz:
- [ ] This is a bug report
- [x] This is a feature request
- [x] I searched existing issues before opening this one

### Expected behavior
Not sure if this is a bug, so I opened as a feature request. I would expect the ignore rules to strip/ignore trailing whitespaces.

### Actual behavior
`# hadolint ignore=DL3029` ignores the rules whereas `# hadolint ignore=DL3029 ` does not.
I know there is no particular reason to have a trailing whitespace at the end of a comment, but it took me a bit to figure it out.

### Steps to reproduce the behavior
Run hadolint on this dockerfile
```Dockerfile
FROM busybox
# hadoling ignore=DL3003
RUN cd /
# hadoling ignore=DL3003 
RUN cd /
```
In my opinion there should be 2 DL3003 warnings, but there is only one.

**Output of `hadolint --version` or
  `docker run --rm hadolint/hadolint hadolint --version` or
  `docker run --rm ghcr.io/hadolint/hadolint hadolint --version`:**

```bash
Haskell Dockerfile Linter 2.12.0
```

**Dockerfile** (if relevant)
see above

avazhnov-twilio:
See also: https://github.com/hadolint/hadolint/issues/1028

PR #1059, opened 2024-11-22
Pragma: Allow comments following inline

m-ildefons:
Allow comments and whitespace to follow inline with pragma comments, e.g.:

```
# hadolint ignore=DL3006  # comment on why to ignore this rule here
# hadolint shell=powershell  
```

Modify the parsers for pragma comments to ignore both:

1) just whitespaces following the inline comment.
2) whitespaces followed by `#`, and anything after until end of line.

This improves robustness of the parser and allows users to add helpful comments to their pragmas.
Note that not separating comment from pragma with an additional `#` will result in a silent failure to parse and the pragma will have no effect.

fixes: hadolint/hadolint#997
fixes: hadolint/hadolint#1028

### How to verify it

These two lines should be two working pragmas, instructing Hadolint to ignore DL3006 on the next Docker instruction and hinting that the shell used in the base image is Powershell (i.e. don't apply rules that are specific to POSIX-like shells).
```
# hadolint ignore=DL3006  # comment on why to ignore this rule here
# hadolint shell=powershell  
```
Test cases are also included in this PR


reteps:
👀 @lorenzo Any way to push this forward?

PR #1070, opened 2025-01-19
Fix resolution for Cabal deps

phansys:
### What I did

I've updated the constraints for the packages with failed resolutions at `cabal.project.freeze`.

### How I did it

I've reviewed the results from the failing install process.

### How to verify it

Check the results of the `Freeze` CI job from the `haskell.yml` workflow.


m-ildefons:
Thanks for the PR, but I'm going to close this since the dependencies have already been bumped to newer versions and this PR is not in conflict.

PR #1074, opened 2025-03-03
Improve DL3009 rule to properly handle all apt-based package managers

### LINKED ISSUE #1073: DL3009 not triggered after `apt update` when `apt` is used instead of `apt-get`
Normally, Hadolint triggers **DL3009** (`Delete the apt-get lists after installing something`) when `/var/lib/apt/lists` is not cleaned up after `apt-get update`. However, when using `apt update` instead of `apt-get update`, DL3009 does not appear. Instead, **DL3027** (`Do not use apt as it is meant to be an end-user tool, use apt-get or apt-cache instead`) is shown.

I'm unsure if this is intentional or if DL3009 does not check for `apt`. Since these two rules address different concerns and should be independent, should DL3009 still be triggered in this case?

Looking forward to your thoughts. Thanks!

PeterDaveHello:
### What I did

Improved DL3009 rule to properly handle all apt-based package managers:
- Added support for apt and aptitude in hasUpdate function
- Updated error message to be more specific about tools and path

### How I did it

1. Enhanced the hasUpdate function to detect update operations from apt and aptitude commands
2. Updated the error message to clearly specify which package managers it applies to and the exact path that needs cleaning

### How to verify it

Create a test Dockerfile with different apt-based package managers:

```dockerfile
FROM ubuntu:20.04

# This will trigger DL3009 (apt-get)
RUN apt-get update && apt-get install -y python3

# This will also trigger DL3009 now (apt)
RUN apt update && apt install -y curl

# This will also trigger DL3009 now (aptitude)
RUN aptitude update && aptitude install -y wget

# This will not trigger DL3009 (proper cleanup)
RUN apt-get update && apt-get install -y git && rm -rf /var/lib/apt/lists/*
```

Run hadolint on the file:
```bash
hadolint Dockerfile
```

The first three RUN instructions should now trigger the DL3009 warning with the updated message.

Fix #1073

### DIFF
--- a/src/Hadolint/Rule/DL3009.hs
+++ b/src/Hadolint/Rule/DL3009.hs
@@ -28,7 +28,7 @@ dl3009 = veryCustomRule check (emptyState Empty) markFailures
   where
     code = "DL3009"
     severity = DLInfoC
-    message = "Delete the apt-get lists after installing something"
+    message = "Delete the apt lists (/var/lib/apt/lists) after installing something"
 
     check line st (From from) = st |> modify (rememberStage line from)
     check line st (Run (RunArgs args flags))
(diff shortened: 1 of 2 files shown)


PeterDaveHello:
We need #1063 to be merged first to pass the CI :sweat_smile:

m-ildefons:
Thanks for the PR.
No idea why the tests didn't pass.
I rebased the commit onto `master` HEAD and ran `cabal test` successfully locally.

PR #1093, opened 2025-06-06
Fix casing in Dockerfile

JEnglandOhalo:
Match casing between FROM and AS statements to address warnings when making a release.

Refs:
- https://github.com/hadolint/hadolint/actions/runs/14010622727/job/39230204603#step:7:212


JEnglandOhalo:
happy to help

PR #1098, opened 2025-07-03
fix(DL3041): Prevent false negatives for packages which include a dash in their package name

### LINKED ISSUE #1002: DNF Rule "Specify Version" Not Applying to dnf-plugins-core
- [x] This is a bug report

### Expected behavior

I would expect a docker line of:

```
RUN dnf install -y dnf-plugins-core \
    && dnf clean all
```

To return an error of:

```
-:5 DL3041 warning: Specify version with `dnf install -y <package>-<version>`.
```

### Actual behavior

Hadolint runs successfully with no errors.

Use this dockerfile:

```docker
FROM rockylinux:9.3
RUN dnf install -y dnf-plugins-core \
    && dnf clean all
```

And run hadolint against it.

This oneliner works:

```
printf "%s\n" "FROM rockylinux:9.3" "RUN dnf install -y dnf-plugins-core && dnf clean all" | docker run --rm -i hadolint/hadolint
```

If you add, e.g., `openssl` to the install command then it works as expected:

```
$ printf "%s\n" "FROM rockylinux:9.3" "RUN dnf install -y openssl dnf-plugins-core && dnf clean all" | docker run --rm -i hadolint/hadolint
-:2 DL3041 warning: Specify version with `dnf install -y <package>-<version>`.
```

```bash
```

**Dockerfile** (if relevant)

```Dockerfile
FROM rockylinux:9.3
RUN dnf install -y dnf-plugins-core \
    && dnf clean all
```

egfx-notifications:
### What I did
Prevent false negatives for packages which include a dash in their package name

### How I did it
Tries to validate presence of a version based on the following assumptions:
- At least one "-" must be present
- All dash-separated parts after the first must contain only valid version characters (according to Fedora Versioning Guidelines)
- At least one dash-separated part after the first must start with a digit

### How to verify it
Unit tests included

Fixes #1002

Please note that this is my first time writing Haskell, so please suggest changes if there are better ways to do this.

### DIFF
--- a/src/Hadolint/Rule/DL3041.hs
+++ b/src/Hadolint/Rule/DL3041.hs
@@ -4,7 +4,8 @@ import qualified Data.Text as Text
 import Hadolint.Rule
 import qualified Hadolint.Shell as Shell
 import Language.Docker.Syntax
-
+import Data.List.Split (splitOn)
+import Data.Char (isDigit)
 
 rule :: Rule Shell.ParsedShell
 rule = dl3041 <> onbuild dl3041
@@ -36,7 +37,29 @@ dnfPackages args =
 
 packageVersionFixed :: Text.Text -> Bool
 packageVersionFixed package =
-  "-" `Text.isInfixOf` package || ".rpm" `Text.isSuffixOf` package
+  let parts = splitOn "-" (Text.unpack package)
+      numParts = length parts
+  in  if numParts <= 1
+      then False  -- No dashes, definitely no version
+      else if ".rpm" `Text.isSuffixOf` package
+           then True -- rpm files always have a version
+           else isVersionLike (map Text.pack $ drop 1 parts)
+
+isVersionLike :: [Text.Text] -> Bool
+isVersionLike parts =
+  case parts of
+    [] -> False  -- No parts after splitting by hyphen
+    _ -> all partIsValid parts && any partStartsWithDigit parts
+  where
+    partIsValid part = Text.all isVersionChar part
+    partStartsWithDigit part = case Text.uncons part of
+                                 Just (c, _) -> isDigit c
+                                 Nothing -> False -- Empty Text
+
+isVersionChar :: Char -> Bool
+isVersionChar c = isDigit c || isAlpha c || c `elem` ['.', '~', '^', '_']
+  where
+    isAlpha char = (char >= 'a' && char <= 'z') || (char >= 'A' && char <= 'Z')
 
 dnfModules :: Shell.ParsedShell -> [Text.Text]
 dnfModules args =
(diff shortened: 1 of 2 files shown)


src/Hadolint/Rule/DL3041.hs
```diff
@@ -36,7 +37,29 @@ dnfPackages args =
...
 packageVersionFixed package =
-  "-" `Text.isInfixOf` package || ".rpm" `Text.isSuffixOf` package
+  let parts = splitOn "-" (Text.unpack package)
+      numParts = length parts
+  in  if numParts <= 1
+      then False  -- No dashes, definitely no version
+      else if ".rpm" `Text.isSuffixOf` package
+           then True -- rpm files always have a version
+           else isVersionLike (map Text.pack $ drop 1 parts)
+
```
m-ildefons, line 47: Nit, just for you : You can avoid nesting conditionals by using pattern matching, e.g. like so:
```
packageVersionFixed :: Text.Text -> Bool
packageVersionFixed package
  | length parts <= 1 = False  -- No dashes, definitively no version
  | ".rpm" `Text.isSuffixOf` package = True  -- rpm files always have a version
  | otherwise = isVersionLike $ drop 1 parts
  where
    parts = Text.splitOn "-" package
```

Additionally, using `Text.splitOn` instead of `Data.List.Split.splitOn` allows you to work without repeatedly packing and unpacking the `Text` data.

src/Hadolint/Rule/DL3041.hs
```diff
@@ -4,7 +4,8 @@ import qualified Data.Text as Text
 import Hadolint.Rule
 import qualified Hadolint.Shell as Shell
 import Language.Docker.Syntax
-
+import Data.List.Split (splitOn)
```
m-ildefons, line 7: This is superfluous when using `Text.splitOn`

src/Hadolint/Rule/DL3041.hs
```diff
@@ -36,7 +37,29 @@ dnfPackages args =
...
+  where
+    partIsValid part = Text.all isVersionChar part
+    partStartsWithDigit part = case Text.uncons part of
+                                 Just (c, _) -> isDigit c
+                                 Nothing -> False -- Empty Text
+
+isVersionChar :: Char -> Bool
+isVersionChar c = isDigit c || isAlpha c || c `elem` ['.', '~', '^', '_']
+  where
+    isAlpha char = (char >= 'a' && char <= 'z') || (char >= 'A' && char <= 'Z')
```
m-ildefons, line 62: This is equivalent to `isAsciiUpper || isAsciiLower` with `isAsciiUpper` and `isAsciiLower` from `Data.Char`.

m-ildefons (approved):
Hi Florian,

thanks for the contribution. It looks pretty good and I'd accept it.

I've put some suggestions, since you asked for it. But don't take them too harshly, they are pretty nit-picky.

egfx-notifications:
> Hi Florian,
> 
> thanks for the contribution. It looks pretty good and I'd accept it.
> 
> I've put some suggestions, since you asked for it. But don't take them too harshly, they are pretty nit-picky.

Thanks! That really helps. I tried to address your feedback with a follow-up commit, please have a look :)

m-ildefons:
Great. Thanks a lot for the contribution.

Issue #1065, opened 2024-12-12
[BUG] hadolint XML checkstyle is wrongly encoded


rantoniuk:
The XML created has wrong entity encoding that causes the UI parsers to show also wrongly encoded information.

Command:
```hadolint -f checkstyle Dockerfile-test```

Output (head):
```
<?xml version='1.0' encoding='UTF-8'?><checkstyle version='4.3'><file name='Dockerfile&#45;test'>
```

Expected:
```
<?xml version="1.0" encoding="UTF-8"?><checkstyle version="4.3"><file name="Dockerfile-test">
```

Resulting wrong UI view after parsing via:
```
    - uses: lcollins/checkstyle-github-action@v3.1.0
      with:
        path: 'checkstyle-hadolint.xml
```

<img width="1261" alt="Screenshot 2024-12-12 at 13 21 18" src="https://github.com/user-attachments/assets/87030347-f1b4-40d7-b94d-bdcdcf5fb504" />

PS. You can [now enable issue types](https://github.com/orgs/community/discussions/112806) to differentiate between Bugs/Features.

PR #1104, opened 2025-08-04
fix character escaping in checkstyle formatter

m-ildefons:
Fix special character escaping in checkstyle formatter. The checkstyle formatter produces an XML document, where only `<`, `>`, `&`, `'` and `"` need to be escaped.

related-to: hadolint/hadolint#1065

### What I did

Fix special character escaping to match XML 1.0 expectation: https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references#Standard_public_entity_sets_for_characters

### How to verify it

Previous to this change:
```xml
<?xml version='1.0' encoding='UTF-8'?><checkstyle version='4.3'><file name='/home/moritz/tmp/Dockerfile' ><error line='3' column='1' severity='warning' message='Specify version with &#96;dnf install &#45;y &#60;package&#62;&#45;&#60;version&#62;&#96;.' source='DL3041' /></file></checkstyle>
```

After this change:
```xml
<?xml version='1.0' encoding='UTF-8'?><checkstyle version='4.3'><file name='/home/moritz/tmp/Dockerfile' ><error line='3' column='1' severity='warning' message='Specify version with `dnf install -y &lt;package&gt;-&lt;version&gt;`.' source='DL3041' /></file></checkstyle>
```

### DIFF
--- a/src/Hadolint/Formatter/Checkstyle.hs
+++ b/src/Hadolint/Formatter/Checkstyle.hs
@@ -74,7 +74,7 @@ toFile results filePathInReport =
     }
   where
     filepath = if null filePathInReport then filename results else getFilePath filePathInReport
-    filename (Result {fileName=fn}) = fn
+    filename Result {fileName=fn} = fn
 
 renderResults ::
   (Foldable f, VisualStream s, TraversableStream s, ShowErrorComponent e) =>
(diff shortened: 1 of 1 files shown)


m-ildefons:
@rantoniuk would this fix your issue?

src/Hadolint/Formatter/Checkstyle.hs
```diff
@@ -78,8 +78,15 @@ escape = Text.concatMap doEscape
     doEscape c =
       if isOk c
         then Text.singleton c
-        else "&#" <> Text.pack (show (ord c)) <> ";"
-    isOk x = any (\check -> check x) [isAsciiUpper, isAsciiLower, isDigit, (`elem` [' ', '.', '/'])]
+        else xmlEscape c
```
rantoniuk, line 81: Maybe you'd like to use [one of the libraries instead](https://mojoauth.com/escaping/xml-escaping-in-haskell/)?

m-ildefons: That's probably overall the best way to go.
I had thought about it but wasn't quite sure which library was best to choose.
I've now decided to go with xml-conduit. It was a quite a bit of work to move the code over, but I'm happy as there is now overall less code and the bits that are there seem less fragile to me.

PR #1111, opened 2025-08-20
added new rule `DL3062` to check `go install`

Danil42Russia:


README.md
```diff
@@ -532,6 +532,7 @@ Please [create an issue][] if you have an idea for a good rule.
 | [DL3059](https://github.com/hadolint/hadolint/wiki/DL3059)   | Info             | Multiple consecutive `RUN` instructions. Consider consolidation.                                                                                    |
 | [DL3060](https://github.com/hadolint/hadolint/wiki/DL3060)   | Info             | `yarn cache clean` missing after `yarn install` was run.                                                                                            |
 | [DL3061](https://github.com/hadolint/hadolint/wiki/DL3061)   | Error            | Invalid instruction order. Dockerfile must begin with `FROM`, `ARG` or comment.                                                                     |
+| [DL3062](https://github.com/hadolint/hadolint/wiki/DL3061)   | Warning          | Pin versions in go install. Instead of `go install <package>` use `go install <package>@<version>`                                                  |
```
larry-c-ef, line 535: This is the wrong address (`https://github.com/hadolint/hadolint/wiki/DL3061`). And theres is no wiki entry for DL3062.

The error message is not helpful enough, especially given that this rule applies to `go get` and `go run`, in addition to `go install`.

albertorm95: `go get -v ./...` makes the lint triggers, so is it not recommended?

Issue #1117, opened 2025-09-02
DL3041: False positives when package version contains a colon


carlsmedstad:
Version 2.13.1 introduced a false positive in the check DL3041 (_Specify version with `dnf install -y <package>-<version>`_) when a version contains a colon, used to declare [package epoch](https://docs.fedoraproject.org/en-US/packaging-guidelines/Versioning/#_epoch_tag) for RPM packages.

Example:

```dockerfile
FROM registry.access.redhat.com/ubi8/ubi-minimal:8.10

# DL3041 triggerd by this:
RUN dnf install --assumeyes \
    jq-1.6 \
    openssl-1:1.1.1k \
    findutils-1:4.6.0 \
    bc-1.07.1 \
    && dnf clean all
```

Removing the colons from the above also removes the DL3041 warning:

```dockerfile
FROM registry.access.redhat.com/ubi8/ubi-minimal:8.10

RUN dnf install --assumeyes \
    jq-1.6 \
    openssl-1.1.1k \
    findutils-4.6.0 \
    bc-1.07.1 \
    && dnf clean all
```

m-ildefons:
Release v2.14.0 is out and includes the linked PR with the fix. I'll close this as completed.

PR #1121, opened 2025-09-03
DL3041, DL3033: Handle RPM package epoch

m-ildefons:
In Fedora, RHEL and similar distributions, packages may have an epoch prefixed to their version. This is an integer separated by a colon from the actual version number. The epoch is used to make all packages sortable, regardless of whether the upstream project uses sortable version identifiers.
To deal with this, in Hadolint we must accept that a dnf or yum package may contain a colon in their version number.

related-to: hadolint/hadolint#1117

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ stack.yaml.lock
 
 #cabal
 dist-newstyle
+dist
 
 # travis wait is generating log in PWD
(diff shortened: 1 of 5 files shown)


m-ildefons:
@carlsmedstad would you like to review this?

Issue #1127, opened 2025-09-18
Bounds incompatible with Stackage LTS


sternenseemann:
In particular, these bounds can be relaxed according to my testing:

```
ilist >=0.3.1 && <0.4, optparse-applicative >=0.19.0 && <0.20
```

m-ildefons:
The latest release doesn't depend on `ilist` anymore and builds with `optparse-applicative >= 0.18.1`, which should work with Stackage 24.10 LTS (GHC 9.10.2).
I'm going to close this as resolved.

PR #1128, opened 2025-09-18
relax dependencies

m-ildefons:
- Relax dependency requirement on optparse-applicative such that it works with stackage 24.10 LTS

- Remove dependency on ilist

related-to: hadolint/hadolint#1127


m-ildefons:
@sternenseemann could you please check if this works for you?

Issue #1130, opened 2025-09-22
New rule DL3062 is overly strict


tkmph:
We use go code (instead of, e.g., shell) to perform builds. We compile and execute the go code from the Dockerfile as follows:
```
RUN go run <path-to-go-file> <arguments>
```

The [new rule DL3062](https://github.com/hadolint/hadolint/pull/1111) flags this use:
```
<error column="1" line="14" message="Pin versions in go. Instead of `go install <package>` use `go install <package>@<version>`" severity="warning" source="DL3062"/>
```

This is overly strict. The go file is managed by version control and matches exactly the other code at the same version. I don't think that the rule should flag invocations of `go run` that execute specific go files (`go run /path/to/file.go`).

PR #1131, opened 2025-09-23
DL3062: relax DL3062, fix readme

m-ildefons:
DL3062 is very eager and sometimes triggers in cases where it doesn't make a whole lot of sense.
To fix this, it now doesn't trigger anymore when installing go packages from local filesystem paths. It also doesn't trigger anymore on command line arguments to go programs that are executed with `go run package@v1.2.3 arg1 arg2`

related-to: hadolint/hadolint#1130


larry-c-ef (commented):
Thanks for tagging me. I don't speak enough Haskell to approve or disapprove, but I commented on the part I know about.

README.md
```diff
@@ -532,7 +532,7 @@ Please [create an issue][] if you have an idea for a good rule.
 | [DL3059](https://github.com/hadolint/hadolint/wiki/DL3059)   | Info             | Multiple consecutive `RUN` instructions. Consider consolidation.                                                                                    |
 | [DL3060](https://github.com/hadolint/hadolint/wiki/DL3060)   | Info             | `yarn cache clean` missing after `yarn install` was run.                                                                                            |
 | [DL3061](https://github.com/hadolint/hadolint/wiki/DL3061)   | Error            | Invalid instruction order. Dockerfile must begin with `FROM`, `ARG` or comment.                                                                     |
-| [DL3062](https://github.com/hadolint/hadolint/wiki/DL3061)   | Warning          | Pin versions in go install. Instead of `go install <package>` use `go install <package>@<version>`                                                  |
+| [DL3062](https://github.com/hadolint/hadolint/wiki/DL3062)   | Warning          | Pin versions in go install. Instead of `go install <package>` use `go install <package>@<version>`                                                  |
```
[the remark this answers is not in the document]
m-ildefons: Thanks for reminding me of that wiki page. I've added it with a variety of example cases, which should make it clear that `go install`, `go get` and `go run` all fall under this rule.

test/Hadolint/Rule/DL3062Spec.hs
```diff
@@ -19,6 +19,28 @@ spec = do
...
+      it "go install with local relative path" $ do
+        ruleCatchesNot "DL3062" "RUN go install ./app/foobar"
+        onBuildRuleCatchesNot "DL3062" "RUN go install ./app/foobar"
+      it "go install local dir" $ do
+        ruleCatchesNot "DL3062" "RUN go install ."
+        onBuildRuleCatchesNot "DL3062" "RUN go install ."
+
+      describe "go install tool" $ do
+        it "version not pinned" $ do
+          ruleCatches "DL3062" "RUN go install tool foobar"
```
medasx, line 35: **issue**: go install treats `tool` argument as package. Only `RUN go install tool` needs to have special handling, as it will install all tools already present in go.mod. 
```
❯ go install tool golang.org/x/tools/cmd/stringer
go: warning: "tool" matched no packages
no required module provides package golang.org/x/tools/cmd/stringer; to add it:
	go get golang.org/x/tools/cmd/stringer
```
Note that tools are installed with `go get -tool [package]` which is already covered by the tests below.

m-ildefons: Thanks for highlighting this. I've removed those test cases.

PR #1148, opened 2025-11-28
Ignore dnf group install from DL3041

kpumuk:
### What I did

`dnf` allows to install groups of packages with `dnf group install`, which do not include version information. This leads to a false positive for

```dockerfile
RUN dnf -y group install "Development Tools" && dnf clean all
```

### How I did it

When `group` is found in the command line, `dnfPackages` list will not be populated with the rest of the arguments.

### How to verify it

See unit tests.

### DIFF
--- a/src/Hadolint/Rule/DL3041.hs
+++ b/src/Hadolint/Rule/DL3041.hs
@@ -30,7 +30,7 @@ dnfPackages :: Shell.ParsedShell -> [Text.Text]
 dnfPackages args =
     [ arg
       | cmd <- Shell.presentCommands args,
-        not (Shell.cmdsHaveArgs dnfCmds ["module"] cmd),
+        not (Shell.cmdsHaveArgs dnfCmds ["module", "group"] cmd),
         arg <- installFilter cmd
     ]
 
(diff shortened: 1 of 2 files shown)


kpumuk:
I would argue that:

* the install procedure is different and there just no way to specify a version with `group install`. If the rule is to be enforced, it should a separate rule like "do not use `dnf group install` or `yum groupinstall` because the version is not specified", so that it can be disabled separately from version requirement
* `yum groupinstall` is not checked for version requirement, so this PR just brings `dnf` to match the logic
* `dnf group install` is often combined with `dnf install pkg-version` to produce a single layer for development dependencies. Separating them into separate docker layers just to disable the version rule for group install does not sound right

guest20:
> the install procedure is different and there just no way to specify a version 

Are you saying `group install` should be exempt from the `you must specify a version` rule because its impossible to specify a version if you use `group install`?

guest20:
> there is no way to differentiate between group install and installing a single package

Isn't this PR to distinguish between `group install` and a regular old `package-version` install? What do you mean there's no way?

kpumuk:
I mean from the hadolint output. Currently, with both `dnf group install` and `dnf install package` the error is the same:

```
DL3041: Specify version with dnf install -y <package>-<version>
```

It is not possible to continue failing on missing version from `dnf install` and silence `dnf group install`, because it is the same rule. This MR recognizes it and removes `group install` out of scope for `DL3041`.

On one hand, the suggestion is simply confusing: since the command is `dnf group install`, and not `dnf install`, there is no way to specify version. A suggestion that cannot be acted on is a bad suggestion in my opinion.

If it instead said something like:

```
DL5555: Do not use `dns group install`
DL5555: Do not use `yum groupinstall`
```

it would:

* Allow users to disable `DL5555` when the `group install` makes sense (for example, installing development tools to compile a library dependency)
* Clearly indicate what is wrong, instead of suggesting to add a version in a context where adding a version is not possible
* Make behavior consistent between `yum groupinstall` (not reported right now) and `dnf group install` (reported with a misleaning suggestion)

m-ildefons (approved):
LGTM, thanks for the PR.

Issue #1150, opened 2025-12-10
Please support file-path-in-report for sonarqube format


304NotModified:
# Feature request

## Actual behaviour

`--file-path-in-report` only works checkstyle format

<img width="658" height="97" alt="Image" src="https://github.com/user-attachments/assets/44df5f62-822b-43c7-9a6e-fbc679d919c6" />
See https://github.com/hadolint/hadolint/blob/master/README.md

SonarQube needs relative paths, and currently we have absolute paths

## Requested 
Support `--file-path-in-report` in sonarqube format, so we could set relative paths

related

https://github.com/hadolint/hadolint/issues/1016

304NotModified:
So I'm a bit confused.

SonarQube gives this warning in the GUI

<img width="524" height="293" alt="Image" src="https://github.com/user-attachments/assets/36de12d1-f5c9-4001-b977-165675de5e4c" />

But the file does exists and the issue is nicely reported in SonarQube - so it does work. 

When changing it to a relative path (using replace in file), the warning is gone but also the issue is not reported in SonarQube... Tried maybe different relative paths and also different paths for the reportfile.

I know this is more a SonarQube issue, but the waning is the reason for this feature request. 

(SonarQube v2025.2)

304NotModified:
Fixed by https://github.com/hadolint/hadolint/pull/1151

PR #1151, opened 2025-12-11
SonarQube Formatter: Support --file-path-in-report

m-ildefons:
Add support for --file-path-in-report flag for SonarQube formatter. This config option lets users set a custom filepath, replacing the real filepath of the Dockerfile in the generated output. This can be used e.g. to hide sensitive file paths when the report is shared with external parties.

related-to: hadolint/hadolint#1150

Screenshot at 2025-12-11 15-21-54


304NotModified (approved):
LGTM :)

Thanks!

Issue #1156, opened 2025-12-12
Feature Request: only check specific rules


304NotModified:
Right now, hadolint checks all rules by default. If you want to use only specific rules, you have to exclude everything else:

```yaml
ignored:
  - DL3000
  - DL3001
  - DL3002
  - DL3003
  - DL3004
  - DL3005
  # TODO ... all rules
```

**Why this is problematic:**
- Every hadolint update adds new rules that can suddenly break CI/CD
- Maintaining a massive exclusion list becomes tedious
- The config doesn't show which rules I actually care about - only what I don't want

## What I'd Like Instead

Just let me specify which rules I **want** to check:

```yaml
only:
  - DL3001
  - DL3007
  - DL3008
```

**Or as alternative: presets**

```yaml
# Start with a sensible baseline
default: recommended  # e.g. all, none, security-focused etc

# Then tweak it
include:
  - DL3025
exclude:
  - DL3007  # We're not ready for this one yet
```

## Why This Would Be Better

- **No more surprise failures**: New rules don't randomly break builds
- **Actually readable config**: 5 lines instead of 50+
- **Clear intent**: Everyone sees what we care about
- **Easy onboarding**: Start with a few rules, add more as we go

304NotModified:
FYI in this validator (https://github.com/304NotModified/SLNX-validator) I had the same issue and I fixed it by allowing *, e.g.

```bash
# Set everything to INFO, but keep SLNX011 as MAJOR
slnx-validator MySolution.slnx --info * --major SLNX011

# Ignore everything except SLNX013
slnx-validator MySolution.slnx --ignore * --major SLNX013
```

PR #1158, opened 2025-12-13
Added DL3063: A USER should be set after the last FROM (default level: ignore)

### LINKED ISSUE #328: DL3002 will pass with the default user as root
Since the default user in docker is root, hadolint should probably complain if the user is never changed to a non-root user.

This fails DL3002:
```
FROM debian:stretch
USER root
CMD ["whoami"]
```

But this does not:
```
FROM debian:stretch
CMD ["whoami"]
```

### LINKED ISSUE #989: Adding a rule to restrict an image to build without adding a non-root user.
This is a feature request to add a hadolint rule, so we can restrict building an image or throw the error while linting an image. This rule will basically check whether in the image user has configure the non-root user or not. As we know, when we build the image and not creating a non-root user then by default image will inherit the root user from base image which is not good from a security point of view. So we should add a rule to restrict the image building if user does not create a non-root user in the image.

https://docs.docker.com/develop/develop-images/instructions/#user

### Expected behavior
Throw the error if user do not configure the non-root user in the Dockerfile.

### Actual behavior
There is no such rule. If user is not creating a non-root user in Dockerfile, still hadolint pass the check.

1. Create a Dockerfile
```
FROM maven:3.8.1-openjdk-17-slim as builder

WORKDIR /build
COPY . /build

RUN mvn dependency:go-offline package -DskipTests=true #remove -DskipTests=true to run the tests while building image

# Step : Package image
FROM openjdk:17-slim
COPY --from=builder /build/target/*.jar /app/app.jar
EXPOSE 9090
#CMD exec java $JAVA_OPTS -jar /app/app.jar
ENV JAVA_OPTS=""
ENTRYPOINT [ "sh", "-c", "java $JAVA_OPTS -jar /app/app.jar ${0} ${@}"
```
2. Run `hadolint Dockerfile`
3. It will pass the check even though there is no command to create a non-root user/group.

### LINKED ISSUE #1025: The USER must be specified
- [ ] This is a bug report

### Expected behavior
When the USER is not specified in a docker file, it should trigger a warning because the user will be root. 
### Actual behavior
The DL3002 warning is checking if the last user is not a root. But if USER is not set the user of the container is root and warning is triggred 

Any docker file not having USER set.

```bash
```

### LINKED ISSUE #1089: Feature Request: Warn if Dockerfile omits USER instruction (defaults to root)
Hi Hadolint team,

Thanks for your excellent work on this tool — it's become a key part of our CI process!

I'd like to request a new rule (or enhancement to DL3002) that triggers a warning when a Dockerfile does not contain any USER instruction. As you're aware, when the USER directive is missing, the container defaults to running as root, which is often an unintended security risk.

**Problem:**
Currently, Hadolint warns when USER root is explicitly specified (via DL3002), but does not warn at all if USER is omitted, even though this implicitly results in root access.

Proposal:
Introduce a new rule (e.g., DL3xxx) or enhance DL3002 to:

Warn when no USER directive is present

Possibly allow suppression via inline comments or CLI flag (as with other rules)

This would help enforce container hardening policies more effectively in security-conscious environments.

Thanks again for your time and consideration!

Best regards,
Sunil

304NotModified:
### What I did
Added a rule that checks of a USER is set after the last FROM. See markdown docs below.

This rule is set to `Ignore` by default as some base images already specify a non-root user, and some build stages intentionally run as root. 

Fixes https://github.com/hadolint/hadolint/issues/1089,
fixes https://github.com/hadolint/hadolint/issues/1025, 
fixes https://github.com/hadolint/hadolint/issues/989, 
fixes https://github.com/hadolint/hadolint/issues/328

closes #1032

### How I did it
With the help of Copilot and thanks to #1032 from @h4kc

### How to verify it
Added tests

image

docs for the wiki:

    # DL3063

    A `USER` should be set after the last `FROM`.

    ## Single-stage Dockerfile

    ### Problematic code:

    ```dockerfile
    FROM busybox
    RUN apt-get update
    CMD ["/app"]
    ```

    ### Correct code:

    ```dockerfile
    FROM busybox
    RUN apt-get update
    USER app
    CMD ["/app"]
    ```

    ## Multi-stage Dockerfile

    ### Problematic code:

    ```dockerfile
    # Builder stage with USER
    FROM node:18 AS builder
    WORKDIR /app
    USER node
    RUN npm install

    # Last stage missing USER - this is the problem!
    FROM node:18-slim
    COPY --from=builder /app /app
    CMD ["node", "app.js"]
    ```

    ### Correct code:

    ```dockerfile
    # Builder stage can run as root
    FROM node:18 AS builder
    WORKDIR /app
    RUN npm install

    # After the last FROM, a USER is required
    FROM node:18-slim
    COPY --from=builder /app /app
    USER node
    CMD ["node", "app.js"]
    ```

    ## Rationale:

    Running containers as root is a security risk. If an attacker gains access to a container running as root, they have full privileges within that container and potentially on the host system.

    By setting a `USER` after the last `FROM`, you ensure the runtime container runs with reduced privileges, following the principle of least privilege. This limits the potential damage if the container is compromised.

    **Important:** This rule only checks if there's a `USER` **after the last `FROM`** in the Dockerfile. In multi-stage builds, earlier build stages are allowed to run as root since they're discarded after the build.

    **Note:** This rule is set to `Ignore` by default as some base images already specify a non-root user. Enable it explicitly if you want to enforce this check:

    ```yaml
    # .hadolint.yaml
    failure-threshold: warning
    override:
    warning:
        - DL3063
    ```


304NotModified:
@m-ildefons could I please get a review? Thanks!

test/RegressionSpec.hs
```diff
@@ -30,6 +30,7 @@ spec = do
               [ "FROM alpine:3",  -- to satisfy DL3061
                 "ARG A_WITHOUT_EQ",
                 "ARG A_WITH_EQ=",
+                "USER app",  -- to satisfy DL3063, even though it is ignored by default
```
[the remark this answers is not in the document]
m-ildefons: This is indeed a bit confusing, but there is an explanation. It's an implementation detail of the `assertChecks` function, which the test suite uses.
The `assertChecks` function hooks directly into `run` from the `Hadolint.Process` module. However, the logic for ignoring rules is applied elsewhere (in the `Hadolint.Lint` module. This means the rules each trigger always, but the linter later decides which rules to ignore or which rules should receive a different severity, based on config settings.


See https://github.com/hadolint/hadolint/blob/fe819dc2d60d8c3876cfa6d783c7b2078da31dc7/src/Hadolint/Lint.hs#L60 ff.

I believe it could be fixed by replacing the call to `Hadolint.Process.run` in `assertChecks` with a call to `Hadolint.Lint.analyze`, but I would have to double check that.

src/Hadolint/Rule/DL3063.hs
```diff
@@ -0,0 +1,34 @@
...
+  = Acc { hasUser :: Bool, reportLine :: Linenumber }
+  | Empty
+  deriving (Eq, Show)
+
+rule :: Rule args
+rule = veryCustomRule check (emptyState Empty) markFailures
+  where
+    code = "DL3063"
+    severity = DLIgnoreC
+    message = "A USER should be set after the last FROM"
```
m-ildefons, line 17: I'm not so sure about making this rule ignored by default.
I think the purpose of this rule would be better served, if it is enabled by default and the message hints at the possibility that the base image uses the `root` user.
It certainly should not be an error level message though, as it will not result in a build failure. A warning or an info would be appropriate, I think.

Most times when new rules are introduced, some people will complain that their CI pipeline now shows an error. But the purpose of a linter is to be noisy and force people to either comply or reason about why it's ok not to comply (and then put an ignore pragma).

m-ildefons (commented):
Thanks for the PR.
In principle this rule is quite good and the PR itself also looks mostly good.

But there are a few edge cases, where the rule will not work ok and I'm not sure how to deal with them.
Consider a Dockerfile with multiple `FROM` statements, but which aren't used as stages building upon each other, but rather as distinct build targets. The Dockerfile for Hadolint itself is a good example here: https://github.com/hadolint/hadolint/blob/master/docker/Dockerfile

The rule would trigger, but only after the last `FROM` statement, despite the fact that there will be docker images produced from all three.

I think the rule should recognize, when a `FROM` statement begins a build stage and when it may be used to produce a docker image.
In the case a `FROM` statement has an alias (`FROM something AS some_alias`) _and_ the alias is referred to again in a later `FROM` statement (`FROM some_alias`), then it can be assumed that it's just a temporary build stage and will be discarded. This isn't guaranteed, but I think it's a reasonable heuristic.
However when a `FROM` statement doesn't have an alias or if the alias isn't referred to in a later `FROM` or `COPY` statement, then it can be assumed that the stage will be used to produce a docker image, regardless of whether it's the last `FROM` statement in the docker file or not.
Note, that you can also copy files from earlier stages using `COPY --from=$IDX`, with `$IDX` being the stage index, starting at 0. So there is a bit of logic required to figure out exacly when a stage is possibly used as temporary build stage, or as build target.

src/Hadolint/Rule/DL3063.hs
304NotModified: ok thanks! I also prefer info or warning :)

 

> Most times when new rules are introduced, some people will complain that their CI pipeline now shows an error.

I'm indeed afraid of that. That was the reason for ignore

304NotModified:
> Thanks for the PR. In principle this rule is quite good and the PR itself also looks mostly good.
> 
> But there are a few edge cases, where the rule will not work ok and I'm not sure how to deal with them. Consider a Dockerfile with multiple `FROM` statements, but which aren't used as stages building upon each other, but rather as distinct build targets. The Dockerfile for Hadolint itself is a good example here: https://github.com/hadolint/hadolint/blob/master/docker/Dockerfile
> 
> The rule would trigger, but only after the last `FROM` statement, despite the fact that there will be docker images produced from all three.
> 
> I think the rule should recognize, when a `FROM` statement begins a build stage and when it may be used to produce a docker image. In the case a `FROM` statement has an alias (`FROM something AS some_alias`) _and_ the alias is referred to again in a later `FROM` statement (`FROM some_alias`), then it can be assumed that it's just a temporary build stage and will be discarded. This isn't guaranteed, but I think it's a reasonable heuristic. However when a `FROM` statement doesn't have an alias or if the alias isn't referred to in a later `FROM` or `COPY` statement, then it can be assumed that the stage will be used to produce a docker image, regardless of whether it's the last `FROM` statement in the docker file or not. Note, that you can also copy files from earlier stages using `COPY --from=$IDX`, with `$IDX` being the stage index, starting at 0. So there is a bit of logic required to figure out exacly when a stage is possibly used as temporary build stage, or as build target.

I will look into this, next week probably

src/Hadolint/Rule/DL3063.hs
304NotModified: > Most times when new rules are introduced, some people will complain that their CI pipeline now shows an error.

Another option for this: https://github.com/hadolint/hadolint/issues/1156

PR #1163, opened 2025-12-23
docs(README): add mise alternate installation documentation

jylenhof:
### What I did

Add mise alternative installation documentation


m-ildefons (commented):
It makes no sense to me to have the entire table of content shifted by one level below the overall heading. Additionally, there isn't really a point in adding the table of content to the table of content.

Otherwise ok.

PR #1167, opened 2026-02-02
Generate GitHub artifact attestations on release

scop:
### What I did

Added support for generating GitHub artifact attestations on release workflow.

Ref https://docs.github.com/en/actions/concepts/security/artifact-attestations

### How I did it

Manually, based on experience with adding it to other projects.

Did _not_ test. But I think this could work.

### How to verify it

Create a release :) Attestations should appear for browsing at https://github.com/hadolint/hadolint/attestations

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -374,6 +374,10 @@ jobs:
       - build-windows
     name: Create GitHub Release
     runs-on: ubuntu-latest
+    permissions:
+      attestations: write
+      contents: write
+      id-token: write
     steps:
 
       - name: Check out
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -374,6 +374,10 @@ jobs:
       - build-windows
     name: Create GitHub Release
     runs-on: ubuntu-latest
+    permissions:
+      attestations: write
+      contents: write
```
m-ildefons, line 379: Going by the docs for the actions/attest Action, this amount of permission isn't needed.
Instead of write permissions to all repo contents, write permissions to artifact-metadata should be enough.
```
artifact-metadata: write
```

m-ildefons (commented):
Thanks for the contribution.

I'll take it to heart and implement attestation.
However as far as I can tell this PR uses too permissive privileges and it doesn't implement any form of attestation for container images.
Therefore I'll close this PR.

.github/workflows/release.yml
scop: `attestations: write` is required for writing attestations.

`contents: write` is required because when declaring job level permissions, the defaults no longer apply but get reset. So without that the job would no longer be able to make the release -- the `Release` step (which is later in the same job) would fail, it requires `contents` write access.

`artifact-metadata: write` is not required for release assets, which this PR is/was about. But if you want to extend it to cover container images, it is required indeed.

Issue #462, opened 2020-09-02
Avoid ENV inside Dockerfile, mainly with sensible data like passwords


mfandrade:
- [ ] This is a bug report
- [x] This is a feature request
- [x] I searched existing issues before opening this one

### Expected behavior

```
$ cat <<EOF > Dockerfile
FROM mariadb:10
ENV MYSQL_ROOT_PASSWORD=my-secret-pw
EOF

$ hadolint Dockerfile
Dockerfile:2 DL???? Explicit env value defined inside Dockerfile
```
OBS: Sensible env vars must be passed to Docker in runtime.
```
$ docker container run -e MYSQL_ROOT_PASSWORD=my-secret-pw ...
```

### Actual behavior

There is no such rule.

### Steps to reproduce the behavior

N/A

**Output of `hadolint --version` or
  `docker run --rm hadolint/hadolint hadolint --version`:**

```bash
Haskell Dockerfile Linter v1.17.4-0-g43bca62
```

**Dockerfile** (if relevant)

N/A

**Additional environment details (OS, stack version, etc.)**

N/A

PR #1188, opened 2026-05-04
New Rule: DL3064 ARG or ENV with sensitive data

m-ildefons:
### What I did

New Rule: DL3064 warn on ARG or ENV with potentially sensitive data. The new rule with warning severity triggers if it detects an ARG or ENV instruction whose name indicates that it might contain sensitive data. The values of the variables in ARG or ENV instructions will be persisted in the container image. This means that if a user puts sensitive data into a variable with an ARG or ENV instruction, that data will be leaked to anyone who has access to the container image.
To avoid this, the ARG and ENV instructions are checked and their names are matched against popular patterns used to inject sensitive data into container images. If such a pattern is identified, the warning is triggered.

Instead of ARG or ENV instructions, it is better to use secret mounts at runtime to inject the sensitive data into the running container. This avoids persisting the sensitive data in the container image. This can also be used during the image building process, e.g. with the `--mount` flag for the RUN instruction.
This is mentioned by Docker in their [best practices documentation](https://docs.docker.com/reference/build-checks/secrets-used-in-arg-or-env/).

related-to: hadolint/hadolint#462

### How I did it

Add new rule and tests.

### How to verify it

Unit test cases included.

The following Dockerfile is an example for what should now be triggering rule DL3064:

```Dockerfile
# this should trigger DL3064
ARG api_token
FROM alpine:3.24

# this should also trigger DL3064
ENV my_password=foobar
```


mikaelol:
This rule is missing from the wiki: https://github.com/hadolint/hadolint/wiki/DL3064 , could you add it?

Issue #1191, opened 2026-05-07
Add JUnit XML as an output format


bcsgh:
- [ ] This is a bug report
- [x] This is a feature request
- [x] I searched existing issues before opening this one

### Request:

A lot of CI environments want JUnit XML files for reporting "test" results and several other lint tools offer that as an export format. Having that added to the list of output formats would be a "really nice to have".

### Alternitives:

The JSON format output could presumable be transmogrified into the right XML, but that seems like an ugly workaround.

### See also:

* https://github.com/hadolint/hadolint/issues/837

Issue #837, opened 2022-05-12
Output result to terminal AND in special format


alexgit2k:
- [] This is a bug report
- [x] This is a feature request
- [x] I searched existing issues before opening this one

### Expected behavior

When integrating Hadolint in Gitlab with the codeclimate-report (see https://github.com/hadolint/hadolint/blob/master/docs/INTEGRATION.md#gitlab-ci) the result is only displayed if
1. a merge-request for the branch exists
**and**
2. the target-branch (of the merge-request) has already a codeclimate-report

If these conditions are not met and the check fails you will not see the errors in the pipeline-output (only in the artefact which needs to be downloaded and is in JSON-Format).

Therefore it would be good, if there are additional options for output-file/format, e.g.:
hadolint --ignore ... Dockerfile **--file reports/hadolint-$CI_COMMIT_SHA.json --file-format gitlab_codeclimate**

### Actual behavior

Output is redirected to a file
```hadolint Dockerfile --no-fail --format gitlab_codeclimate > reports/hadolint-$CI_COMMIT_SHA.json```

### Steps to reproduce the behavior

See https://github.com/hadolint/hadolint/blob/master/docs/INTEGRATION.md#gitlab-ci

### Workaround

Run hadlint twice:
```
  script:
    - mkdir -p reports
    - hadolint --ignore ... Dockerfile --no-fail --format gitlab_codeclimate > reports/hadolint-$CI_COMMIT_SHA.json
    - hadolint --ignore ... Dockerfile
```

**Comment by xlr-8** [comment]
You could also use `tee` to output it into several destinations without needing to re-run hadolint. That implies that `tee` binary is installed on the server/docker though, now that would also output JSON into your terminal - which isn't ideal either so, you could also use `jq .`

```bash
>  echo '{"foo": "bar"}' | tee -a file.log
{"foo": "bar"}
> cat file.log
{"foo": "bar"}
> jq . file.log
{
  "foo": "bar"
}
```

I know it doesn't quite solve your issue, but figured you might eventually save a bit of time instead of running it several times; and those might be of interested for those who don't know them.

alexgit2k:
Interesting idea!

The two hadolint-runs only take 6 seconds including docker-start. So I think it's a lot of more work to find out the right parameters for jq to get near the default format output of hadolint.

nejch:
[...]

PR #1192, opened 2026-05-13
Add JUnit output formatter

m-ildefons:
### What I did

Add output formatter for JUnit report format. This report format is popular and well integrated with several CI solutions (e.g. Jenkins).

JUnit XML schema, fields and their values are analogue to https://www.ibm.com/docs/en/developer-for-zos/17.0.x?topic=formats-junit-xml-format

Example:
Screenshot at 2026-05-13 22-19-21

related-to: hadolint/hadolint#1191
related-to: hadolint/hadolint#837

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,9 +169,9 @@ Available options:
   --file-path-in-report FILEPATHINREPORT
                            The file path referenced in the generated report.
                            This only applies for the 'checkstyle', 'codeclimate',
-                           'sonarqube' and 'gitlab_codeclimate' formats and is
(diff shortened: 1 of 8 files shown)


m-ildefons:
@bcsgh could you please take a look if this is what you're looking for?

bcsgh:
Based on the screenshot, I think that would work.

(Though for non technical reasons I'll likely have to wait for #837 before I can actually start using it.)

mat813:
That looks exactly like what I was looking for.

Issue #1096, opened 2025-07-02
Allow to ignore certain rules for an entire stage


felipecrs:
It would be great if I could ignore certain rules for an entire stage with something like `# hadolint stage ignore`. Example:

```dockerfile
# hadolint stage ignore=DL3004
FROM somebaseimage AS builder

RUN sudo touch /etc/someconfig.txt

FROM alpine

COPY --from=builder /etc/someconfig.txt /etc/someconfig.txt
```

It would be safer than using `# hadolint global ignore`.

m-ildefons:
Fixed in v2.15.0

PR #1195, opened 2026-05-19
Ignore Pragma: Add stage ignore pragma

m-ildefons:
### What I did

Add stage ignore pragma. This works like the global ignore pragma, but only applies to the stage immediately following the pragma comment. Stage ignore pragmas that aren't immediately followed by a FROM instruction will be silently ignored.

The finer granularity of this pragma comment allows certain rules to be ignored e.g. on build stages and then later applied to the final stages of the image.

related-to: hadolint/hadolint#1096

### How to verify it

The following dockerfile is an example of the properties of the stage ignore pragma

```dockerfile
FROM debian:foobar

EXPOSE 80000  # <-- this should give a warning

# hadolint stage ignore=DL3011
FROM debian:foobar

EXPOSE 80000  # <-- the warning here should be ignored, no output for this line

FROM debian:foobar

EXPOSE 80000  # <-- this should give a warning again
```

### DIFF
--- a/src/Hadolint/Pragma.hs
+++ b/src/Hadolint/Pragma.hs
@@ -1,5 +1,6 @@
 module Hadolint.Pragma
   ( ignored,
+    stageIgnored,
     globalIgnored,
     parseIgnorePragma,
     parseShell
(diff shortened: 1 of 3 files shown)


lorenzo (approved):
looks good!

PR #1196, opened 2026-05-19
Fix DL3013 false positive on pipx --python flag

### LINKED ISSUE #1067: bug(pin-versions-in-pip): false positive when using pipx DL3013
- [x] This is a bug report

### Expected behavior

The following is part of my Dockerfile
```Dockerfile
ENV POETRY_VERSION=1.8.5
RUN pipx install --python "$(which python)" "poetry==$POETRY_VERSION" 
```

It should not trigger any linting fails

### Actual behavior

The following warning is issued
```
1. hadolint: Pin versions in pip. Instead of `pip install <package>` use `pip install <package>==<version>` or `pip install --requirement <requirements file>` [DL3013]
```

I'm using mason in neovim so this is the only output i have.
```bash
    ◍ hadolint
      Dockerfile linter, validate inline bash, written in Haskell.

      installed version v2.12.0                              
      homepage          https://github.com/hadolint/hadolint 
      languages         Docker                               
      categories        Linter                               
      executables       hadolint                             

```

I'm using Arch btw. Neovim too

eyupcanakman:
### What I did

DL3013 reports a missing version pin for `RUN pipx install --python "$(which python)" "poetry==1.8.5"`, because it reads the argument of the `--python` flag as a package. `poetry` itself is pinned.

### How I did it

Added `--python` to the list of flags whose argument DL3013 drops before it builds the package list. DL3013 still applies to `pipx install`, the same as `pip install`.

### How to verify it

`cabal test`

fixes #1067

### DIFF
--- a/src/Hadolint/Rule/DL3013.hs
+++ b/src/Hadolint/Rule/DL3013.hs
@@ -80,6 +80,7 @@ packages cmd =
           "prefix",
           "progress-bar",
           "proxy",
+          "python",
           "python-version",
           "root",
           "src",
--- a/test/Hadolint/Rule/DL3013Spec.hs
+++ b/test/Hadolint/Rule/DL3013Spec.hs
@@ -168,3 +168,9 @@ spec = do
     it "pipenv is not pip" $ do
       ruleCatchesNot "DL3013" "RUN pipenv install black"
       onBuildRuleCatchesNot "DL3013" "RUN pipenv install black"
+    it "pipx version not pinned" $ do
+      ruleCatches "DL3013" "RUN pipx install black"
+      onBuildRuleCatches "DL3013" "RUN pipx install black"
+    it "pipx install --python argument is not a package" $ do
+      ruleCatchesNot "DL3013" "RUN pipx install --python \"$(which python)\" \"poetry==1.8.5\""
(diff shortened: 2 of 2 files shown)


m-ildefons:
Hi,

thanks for the contribution.

I don't think this is the right solution.
`pipx` uses `pip` under the hood, it just adds automatic management for virtual environments on top.
Therefore rule DL3013 can also be applied to `pipx` in the same way. Excluding `pipx` in general here isn't a good solution I think.

If version pinning doesn't make sense for your use case, you should ignore the rule (see the readme file for options how to do that).

However https://github.com/hadolint/hadolint/issues/1067 is indeed a valid bug report.
The problem is that with this input:
```dockerfile
RUN pipx install --python "$(which python)" "poetry==1.8.5"
```
Hadolint thinks that `"$(which python)"` is also a package to install. Since there is no version constraint given, it raises the warning.

However `"$(which python)"` is actually an argument to the `--python` flag for the `pipx install` command.
This doesn't need a version constraint.

There is a list of flags whose arguments to drop. Adding the `--python` flag there will solve the problem correctly.

eyupcanakman:
Thanks, that makes sense. I've updated the PR to add `--python` to the drop-flag list instead, so an unpinned `pipx install` still triggers DL3013.

m-ildefons (approved):

PR #1197, opened 2026-05-20
Add output to file/multiple files

m-ildefons:
### What I did

Add command line option to write output directly to a file. This makes it possible to use Hadolint in scenarios where the output can not be redirected by a command shell or multiple output destinations are needed.

The new command line option is `-o` or `--output`. It can be given multiple times, in which case the result is written to all destinations.

The default behavior remains the same, Hadolint will still print the results on stdout by default.

The result can also be forced on stdout when using the special file name `-`, e.g.: `hadolint -o - `. This is usefule when writing the output to multiple destinations.

This changes the interfaces of the formatters. Instead of a `printResults` function, which always writes to stdout, the formatters now expose an `hWrite` function, which takes a file handle as additional argument.
This decouples the output destination from the formatters and lets Hadolint call the formatters multiple times to write to multiple output destinations.
It also lets Hadolint choose where to write the result based on command line options, rather than having the destination hard-coded in the formatters themselves.

Additionally, the `-f/--format` option is now accepted multiple times. Each format option will be applied to the corresponding output, i.e. the first format to the first output, the second format to the second output and so on.
Each `-f/--format` without a corresponding `-o/--output` option will be printed to the default destination: stdout. Each output without a corresponding `-f/--format` will be printed in the default format.
The formats can also be set via the config file with the new key `formats`, e.g.:
```yaml
formats:
  - json
  - sarif
```
Or via the environment variable `HADOLINT_FORMAT` as comma separated list.
The existing config file key `format` (without `s` at the end) is still valid and will be taken into account, i.e.
```yaml
format: tty
formats:
  - json
  - sarif
```
is the same as `-f tty -f json -f sarif`.

A practical example would be the following invocation, producing human readable output on the command line stdout as well as a machine readable report file:
```
hadolint -f junit -o report.xml -f tty Dockerfile
```

related-to: hadolint/hadolint#837

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -35,8 +35,8 @@ exitProgram ::
   IO ()
 exitProgram conf res
   | noFail conf = exitSuccess
-  | CodeclimateJson == format conf = exitSuccess
-  | Codacy == format conf = exitSuccess
+  | CodeclimateJson `elem` formats conf = exitSuccess
+  | Codacy `elem` formats conf = exitSuccess
   | all (`noFailure` failureThreshold conf) res = exitSuccess
   | otherwise = exitFailure
 
@@ -49,7 +49,7 @@ runLint cmd conf = do
       filePathInReport = filePathInReportOption cmd
       destinations = output cmd
   res <- Hadolint.lintIO conf files
-  write destinations (format conf) (noColor conf) filePathInReport res
+  write destinations (formats conf) (noColor conf) filePathInReport res
   exitProgram conf res
 
 execute :: CommandlineConfig -> Configuration -> IO ()
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -2,7 +2,7 @@ module Main where
 
 import Control.Monad (when)
 import Data.Default
-import Hadolint (OutputFormat (..), printResults, DLSeverity (..))
+import Hadolint (OutputFormat (..), write, DLSeverity (..))
 import Hadolint.Config
 import Prettyprinter
 import qualified Data.List.NonEmpty as NonEmpty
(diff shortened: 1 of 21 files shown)


bcsgh:
The PR description doesn't [edit: didn't] speak to if the different destinations can/can't be different formats.

Is that being added here or is that being left as follows on work? If the second, it would likely be desirable to at least sketch what that will look like to ensure the flags being added/changed here won't need to get incompatible changes there.

m-ildefons:
As it is right now, this PR only implements output to one or multiple destinations.
The output format remains a global setting applicable to all output destinations.

Outputs in different formats to different destinations would be an additional change on top of this.
I'm thinking the most canonical way to implement that would be to change the `-f/--format` flag such that it can be given multiple times. Then the first instance of the flag would set the format for the first output destination and so on.
In case there are more output destinations that `-f/--format` flags, the default output format can be used for the last few output destinations. In case there are more `-f/--format` flags, the last few format flags will be ignored.
E.g.:
```
hadolint -o myreport.log -f sarif -o myjsonreport.json -f json -o - Dockerfile
```
would write `myreport.log` in SARIF format, `myjsonreport.json` in JSON format and to the command line stdout in TTY format.
IMO this would result in a fairly intuitive and logical command line interface. As far as I can tell, all existing invocations would keep their current behavior this way.

bcsgh:
Another option would be to have each `--output` respect the most recent `--format`. That would be potentially less confusing if someone did `-f ... -f ... -o ... -f ... -o ... -o ...` or the like.

m-ildefons:
> Another option would be to have each --output respect the most recent --format.

This approach would result in silently dropping `--format` options. I think that could lead to some confusing situations.
I'd rather have the behavior be noisy enough to see what's happening, i.e. if there are more `--format` options, then just output to stdout.

I also don't expect many people to do crazy complicated combinations of `-o`s and `-f`s.
But people will want their existing invocations to keep their behaviors.

bcsgh:
I think the backwards compatibility concern is a valid point.

I think the risk of confusion is at best a wash as both patterns can be confusing in some cases.

bcsgh:
Poke?

PR #1198, opened 2026-05-24
Fix DL3013 false positive on pip --root-user-action flag

### LINKED ISSUE #1064: DL3013: argument mistaken for package
- [x] This is a bug report

When setting `--root-user-action ignore` in the pip install command `ignore` is interpreted as a Python package and Hadolint wants me to pin a version for it. `ignore` is a argument and should not be mistaken for a package.

```bash
Dockerfile:3 DL3013 warning: Pin versions in pip. Instead of `pip install <package>` use `pip install <package>==<version>` or `pip install --requirement <requirements file>`
```

**Version**

```bash
Haskell Dockerfile Linter 2.12.0
```

**Dockerfile**

```Dockerfile
FROM python:3.12.8-alpine3.20

RUN pip install --no-cache-dir --root-user-action ignore poetry==1.8.5
```

```bash
ProductName:		macOS
ProductVersion:		14.5
BuildVersion:		23F79
Darwin Kernel Version 23.5.0: Wed May  1 20:12:58 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T6000
```

eyupcanakman:
### What I did

DL3013 fires on `RUN pip install --no-cache-dir --root-user-action ignore poetry==1.8.5`. It reads `ignore`, the value of `--root-user-action`, as a package with no version pin. `poetry` is already pinned.

### How I did it

Added `--root-user-action` to the list of value-taking flags the rule drops. On its own that broke the `=` form, where `dropFlagArg` dropped the package instead of the attached value. Now `dropFlagArg` keeps the next argument when the value is attached with `=`.

### How to verify it

`cabal test`

fixes #1064

### DIFF
--- a/test/Hadolint/ShellSpec.hs
+++ b/test/Hadolint/ShellSpec.hs
@@ -51,3 +51,8 @@ spec =
                         [ CmdPart "loglevel" 1]
        in do
         dropFlagArg ["loglevel"] cmd `shouldBe` res
+  --
+    it "dropFlagArg keeps the next argument when the flag value is inline" $ do
+      let cmds = presentCommands (parseShell "npm --loglevel=info install bla@1.0.0")
+      cmds `shouldNotBe` []
+      map (dropFlagArg ["loglevel"]) cmds `shouldBe` cmds
--- a/src/Hadolint/Rule/DL3013.hs
+++ b/src/Hadolint/Rule/DL3013.hs
@@ -83,6 +83,7 @@ packages cmd =
           "python",
           "python-version",
           "root",
+          "root-user-action",
           "src",
           "t",
           "target",
--- a/src/Hadolint/Shell.hs
+++ b/src/Hadolint/Shell.hs
@@ -219,7 +219,10 @@ hasArg arg Command {arguments} = not $ null [a | CmdPart a _ <- arguments, a == 
 dropFlagArg :: [Text.Text] -> Command -> Command
 dropFlagArg flagsToDrop Command {name, arguments, flags} = Command name filteredArgs flags
   where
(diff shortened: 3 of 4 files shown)


test/Hadolint/ShellSpec.hs
```diff
@@ -51,3 +51,12 @@ spec =
                         [ CmdPart "loglevel" 1]
        in do
         dropFlagArg ["loglevel"] cmd `shouldBe` res
+  --
+    it "dropFlagArg keeps the next argument when the value is inline" $ do
+      let cmd = Command "npm"
+                        [ CmdPart "--loglevel=info" 1,
+                          CmdPart "install" 2,
+                          CmdPart "bla@1.0.0" 3]
+                        [ CmdPart "loglevel" 1]
```
m-ildefons: This structure is incorrect. It should not have multiple `CmdPart "..." 1`, as the index for each `CmdPart` object should be unique.

```
npm --loglevel=info install bla@1.0.0
```
would parse into
```
Command "npm"
  [ CmdPart "install" 2,
    CmdPart "bla@1.0.0" 3 ]
  [ CmdPart "loglevel=info" 1 ]
```

m-ildefons (commented):
Thanks for the contribution.

The change to DL3013 and the `dropFlagArg` function look good to me. Only the test case seems to be off, since the `Command` structure used in that test case is not what the parser would produce.

test/Hadolint/ShellSpec.hs
eyupcanakman: Switched the test to build the command from `parseShell`.

m-ildefons (approved):
Thanks. Looks good now.

PR #1206, opened 2026-06-10
Update to Alpine 3.24

jnoordsij:
### What I did

Bump to the latest Alpine version, 3.24. See also https://alpinelinux.org/posts/Alpine-3.24.0-released.html.


docker/Dockerfile
m-ildefons: May as well bump the Debian base image.

```suggestion
FROM debian:trixie-slim AS debian
```

m-ildefons (approved):
Thanks for the PR.
LGTM in general.

docker/Dockerfile
jnoordsij: Ah yeah, fair :+1:

m-ildefons (approved):

Issue #1208, opened 2026-06-17
Cover `HEALTHCHECK` `CMD` in DL3025


scop:
Like `CMD` and `ENTRYPOINT`, `HEALTHCHECK`'s `CMD` benefits from being specified as an array rather than a simple command. With it, the array form avoids inadvertent use of shell which might have a tiny performance effect, as well as can pave way toward being able to be based off a minimal image such as [distroless](https://github.com/GoogleContainerTools/distroless) or `scratch`.

https://docs.docker.com/reference/dockerfile/#healthcheck

I think checking this would conceptually fit well in DL3025.

m-ildefons:
Sounds quite reasonable to me.
Would you like to take a peek at the PR and tell me if that's what you had in mind?

scop:
It does conceptually yes. Left one note re arguments before CMD.

m-ildefons:
Fixed in v2.15.0

PR #1209, opened 2026-06-19
DL3025: Trigger on HEALTHCHECK instructions too

m-ildefons:
### What I did

Make DL3025 trigger on HEALTHCHECK instructions too. In some containers based on Distroless or Scratch images, there is no shell, but that doesn't necessarily mean that there can't be a HEALTHCHECK instruction.
To avoid the overhead of a shell, HEALTHCHECK instructions should utilize the JSON notation to directly execute the command without the implicit shell.

related-to: hadolint/hadolint#1208

### How to verify it

This is now ok and won't trigger rule DL3025:
```dockerfile
FROM scratch

HEALTHCHECK CMD [ "foobar" ]
```

This will trigger a warning for rule DL3025:
```dockerfile
FROM scratch

HEALTHCHECK CMD foobar
```

### DIFF
--- a/src/Hadolint/Rule/DL3025.hs
+++ b/src/Hadolint/Rule/DL3025.hs
@@ -12,5 +12,6 @@ rule = simpleRule code severity message check
 
     check (Cmd (ArgumentsText _)) = False
     check (Entrypoint (ArgumentsText _)) = False
+    check (Healthcheck (Check (CheckArgs (ArgumentsText _) _ _ _ _ _))) = False
     check _ = True
 {-# INLINEABLE rule #-}
--- a/test/Hadolint/Rule/DL3025Spec.hs
+++ b/test/Hadolint/Rule/DL3025Spec.hs
@@ -11,24 +11,28 @@ spec = do
   let ?config = def
 
   describe "DL3025 - Use arguments JSON notation for `CMD` and `ENTRYPOINT` arguments" $ do
+
     it "warn on ENTRYPOINT" $
       let dockerFile =
             [ "FROM node as foo",
               "ENTRYPOINT something"
             ]
        in ruleCatches "DL3025" $ Text.unlines dockerFile
+
     it "don't warn on ENTRYPOINT json notation" $
       let dockerFile =
             [ "FROM scratch as build",
               "ENTRYPOINT [\"foo\", \"bar\"]"
             ]
        in ruleCatchesNot "DL3025" $ Text.unlines dockerFile
+
     it "warn on CMD" $
       let dockerFile =
             [ "FROM node as foo",
               "CMD something"
             ]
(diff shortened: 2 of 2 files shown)


src/Hadolint/Rule/DL3025.hs
```diff
@@ -12,5 +12,6 @@ rule = simpleRule code severity message check
 
     check (Cmd (ArgumentsText _)) = False
     check (Entrypoint (ArgumentsText _)) = False
+    check (Healthcheck (Check (CheckArgs (ArgumentsText _) _ _ _ _ _))) = False
```
scop, line 15: I'm a bit of a Haskell illiterate, but does this work properly when there are some options before CMD?

To use an example [from the Dockerfile reference](https://docs.docker.com/reference/dockerfile/#healthcheck), does it warn properly about this:

```
HEALTHCHECK --interval=30s \
  CMD curl -f http://localhost/ || exit 1
```

...and does it properly **not** warn about this:

```
HEALTHCHECK --interval=30s \
  CMD ["sh", "-c", "curl -f http://localhost/ || exit 1"]
```

m-ildefons: Screenshot at 2026-06-22 09-55-50

Seems to work alright, I'd say.

The `check` function does [pattern matching](https://learnyouahaskell.github.io/syntax-in-functions.html) on [dockerfile Instruction](https://github.com/hadolint/language-docker/blob/e262ffee24155ef5f4cde5f1609f3b4f0c56d0e4/src/Language/Docker/Syntax.hs#L395) constructors.
A Dockerfile text gets parsed into a syntax tree (a list of dockerfile Instructions) by the parser from the language-docker library. Then Hadolint evaluates the rule on each of the dockerfile Instructions.
The order of option flags is lost during parsing, so

```
HEALTHCHECK \
  --timeout=30s \
  --interval=30s \
  CMD ["sh", "-c", "curl -f http://localhost/ || exit 1"]
```

and

```
HEALTHCHECK \
  --interval=30s \
  --timeout=30s \
  CMD ["sh", "-c", "curl -f http://localhost/ || exit 1"]
```

have the same representation in the syntax tree.

Here with this rule, if an instruction matches the `Healthcheck` constructor and the check arguments match the `ArgumentsText` constructor (i.e. the commands are given in shell-notation), then the check fails. The flag options for the Healthcheck instruction are ignored in the pattern matching, as indicated by the `_` characters for the [`CheckArgs` constructor](https://github.com/hadolint/language-docker/blob/e262ffee24155ef5f4cde5f1609f3b4f0c56d0e4/src/Language/Docker/Syntax.hs#L259).
Therefore the option flags don't matter, the rule triggers always if the `HEALTHCHECK CMD ` is given in shell-notation.

I hope this explanation helps.

PR #1224, opened 2026-07-30
fix(release): use macos-15-intel runner for x86_64 macOS builds

### LINKED ISSUE #1223: v2.15.0: macOS x86_64 binary is actually an arm64 binary
## Summary

The v2.15.0 release ships an **arm64 binary** under the name `hadolint-macos-x86_64`. This means the binary cannot run on Intel (x86_64) Macs.

## Evidence

| File | Actual architecture | Size | SHA256 |
|---|---|---|---|
| `hadolint-macos-arm64` | arm64 | 102,590,520 | `1d701a41d2760ec88f8eff8d93f78b743d6ed57c53146ef274dbeb540816e563` |
| `hadolint-macos-x86_64` | **arm64** (same!) | 102,590,520 | `1d701a41d2760ec88f8eff8d93f78b743d6ed57c53146ef274dbeb540816e563` |

```shell
$ file hadolint-macos-x86_64
hadolint-macos-x86_64: Mach-O 64-bit executable arm64

$ lipo -info hadolint-macos-x86_64
Non-fat file: hadolint-macos-x86_64 is architecture: arm64

$ diff hadolint-macos-arm64 hadolint-macos-x86_64
# (identical - no output)
```

For comparison, v2.14.0's x86_64 binary was correctly built:

```shell
$ file hadolint-v2.14.0-macos-x86_64
hadolint-v2.14.0-macos-x86_64: Mach-O 64-bit executable x86_64
```

## Impact

Running the binary on Intel Macs produces:

```
Bad CPU type in executable
```

## Expected behavior

`hadolint-macos-x86_64` should be a **x86_64** (or universal/fat) binary that runs on Intel Macs.

shenxianpeng:
macos-15 runner is Apple Silicon (ARM64), so both x86_64 and arm64 builds produce arm64 binaries. Use macos-15-intel for the x86_64 target so it actually compiles for Intel Macs.

Fixes #1223.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -176,7 +176,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - os: macos-15
+          - os: macos-15-intel
             target: macos-x86_64
           - os: macos-15
             target: macos-arm64


m-ildefons:
Thanks for the PR.
However, I think the problem goes a bit deeper, because aside from accidentally building for the same architecture twice, the build artifacts turned out to be the exact same.
If you dig into the logs of the builds, you'll find that both runners pulled from the same cache, because the keys by which the build cache is managed did not include the architecture.
I've opened another PR to fix that as well.

PR #1226, opened 2026-07-30
DL3066: Fix false positive when UID is an argument

m-ildefons:
### What I did

Fix possible false positive when the UID for the USER instruction is given as an argument with an ARG instruction.
In this case, the UID will be injected build-time and Hadolint can't tell if it's all-numeric or not.
It's then best not to alert.

### How to verify it

This should no longer get flagged with DL3066:

```dockerfile
ARG APP_UID
FROM foobar:barfoo

USER ${APP_UID}
```

### DIFF
--- a/src/Hadolint/Rule/DL3066.hs
+++ b/src/Hadolint/Rule/DL3066.hs
@@ -1,19 +1,31 @@
 module Hadolint.Rule.DL3066 (rule) where
 
 import qualified Data.Char as Char
+import qualified Data.Set as Set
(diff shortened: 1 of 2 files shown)


damian-buho:
Dockerfile supports reading USER from ENVs in base image too.
If base image has UID=1000, you can use USER ${UID} in downstream Dockerfile.
It helps to keep uids DRY in big collections, so I would suggest to ignore all ${VAR} expressions, even those without clear ARG/ENV.

PR #1231, opened 2026-08-01
DL4006: match shell by basename, not full exact path

### LINKED ISSUE #977: DL4006 is triggered when using non-bash/ash shells
- [x] This is a bug report

### Expected behavior

A valid Containerfile lints successfully.

### Actual behavior

I get a big fat red exclamation point in my CI.

```bash
❯ podman run -t -v ../auspex:/foo -w /foo --rm docker.io/hadolint/hadolint hadolint --version
❯ podman run -t -v ../auspex:/foo -w /foo --rm docker.io/hadolint/hadolint hadolint Dockerfile
Dockerfile:4 DL4006 warning: Set the SHELL option -o pipefail before RUN with a pipe in it. If you are using /bin/sh in an alpine image or if your shell is symlinked to busybox then consider explicitly setting your SHELL to /bin/ash, or disable this check
```

**Dockerfile** (if relevant)

```Dockerfile
FROM docker.io/tensorflow/tensorflow:latest-gpu
WORKDIR /usr/local/src
SHELL ["/bin/sh", "-o", "pipefail", "-c"]
RUN curl -fsSL http://example.com/foo.tar.gz | tar zx
```

This Dockerfile lints fine:

```Dockerfile
FROM docker.io/tensorflow/tensorflow:latest-gpu
WORKDIR /usr/local/src
SHELL ["/bin/bash", "-o", "pipefail", "-c"]
RUN curl -fsSL http://example.com/foo.tar.gz | tar zx
```

```shell
$ podman run -t -v ../auspex:/foo -w /foo --rm docker.io/hadolint/hadolint hadolint Dockerfile
$
```

See? Nothing!

ankit090701:
### What I did

Fixed DL4006 false-positives when the SHELL directive names a shell via a path other than the exact hardcoded strings `/bin/bash`, `/bin/zsh`, `/bin/ash`, `bash`, `zsh`, `ash` - for example `/usr/bin/bash` (common on distros where `/bin` is a symlink to `/usr/bin`, as in the linked issue):

```Dockerfile
FROM docker.io/tensorflow/tensorflow:latest-gpu
WORKDIR /usr/local/src
SHELL ["/usr/bin/bash", "-o", "pipefail", "-c"]
RUN curl -fsSL http://example.com/foo.tar.gz | tar zx
```

Previously this incorrectly triggered:
```
DL4006 warning: Set the SHELL option -o pipefail before RUN with a pipe in it. ...
```
even though `-o pipefail` is set correctly.

### How I did it

`hasPipefailOption` in `src/Hadolint/Rule/DL4006.hs` required an exact string match of the SHELL command's `name` against a small fixed list of complete strings. Any other valid path pointing at bash/zsh/ash (not just `/usr/bin/bash` - any prefix, really) failed to match.

Added `Shell.shellBasename` in `src/Hadolint/Shell.hs`, which takes the last `/`-separated segment of a command name (so `/bin/bash`, `/usr/bin/bash`, and a bare `bash` all resolve to `"bash"`), and match against that instead of the full path. This only changes *how* the shell name is matched - which shells are considered pipefail-capable (bash/zsh/ash) is unchanged, so `/bin/sh` still correctly triggers the warning (sh/dash doesn't support `-o pipefail`).

I didn't touch the existing (separate, unexported) `basename` helper in `DL3010.hs` - that one operates on `ADD`/`COPY` destination file paths, a different domain from shell command names, even though the string-splitting logic looks similar.

### How to verify it

- `cabal test` - all 744 examples pass (0 failures), including 15 in `DL4006Spec` (13 existing + 2 new)
- Added two regression test cases to `test/Hadolint/Rule/DL4006Spec.hs`: `/usr/bin/bash` and `/usr/local/bin/zsh` with `-o pipefail` set should not warn
- Manually verified against the exact Dockerfile from the issue with a built `hadolint` binary: exits 0, no DL4006 warning
- Manually verified the negative case still works: `SHELL ["/bin/sh", "-o", "pipefail", "-c"]` still correctly triggers DL4006 (sh doesn't support pipefail)

Fixes #977

### DIFF
--- a/src/Hadolint/Rule/DL4006.hs
+++ b/src/Hadolint/Rule/DL4006.hs
@@ -32,8 +32,7 @@ rule = customRule check (emptyState False)
         null
           [ True
             | cmd@(Shell.Command name arguments _) <- Shell.presentCommands script,
-              validShell <- ["/bin/bash", "/bin/zsh", "/bin/ash", "bash", "zsh", "ash"],
-              name == validShell,
+              Shell.shellBasename name `elem` ["bash", "zsh", "ash"],
               Shell.hasFlag "o" cmd,
               arg <- Shell.arg <$> arguments,
               arg == "pipefail"
--- a/src/Hadolint/Shell.hs
+++ b/src/Hadolint/Shell.hs
@@ -114,6 +114,14 @@ shellcheck (ShellOpts sh env) (ParsedShell txt _ _) =
 nonPosixShells :: [Text.Text]
 nonPosixShells = ["pwsh", "powershell", "cmd"]
 
+-- | Returns the last path segment of a shell command name, so that
+-- "/usr/bin/bash" and "/bin/bash" both resolve to "bash" just like a bare
+-- "bash" would. This lets shell-name matching (e.g. detecting bash/zsh/ash
+-- for the "-o pipefail" check) work regardless of which absolute path a
+-- shell happens to be installed under.
+shellBasename :: Text.Text -> Text.Text
+shellBasename = last . Text.splitOn "/"
+
 hasUnsupportedShebang :: Text.Text -> Bool
 hasUnsupportedShebang script =
   "#!" `Text.isPrefixOf` script &&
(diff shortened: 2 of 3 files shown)


m-ildefons (approved):
LGTM, thanks.

PR #1238, opened 2026-08-10
Hadolint Online: Fix GHC WASM install

m-ildefons:
Fix GHC WASM install in Hadolint Online build pipeline
See also:
https://gitlab.haskell.org/ghc/ghc/-/work_items/27577


lorenzo:
amazing!

